PR #115, opened 2015-01-02
Mnist extraction source

subutai:


breznak:
@subutai you da real MVP! :trophy:

PR #136, opened 2015-02-16
Implements simulation code using unions

### LINKED ISSUE #133: Implement SDR calculations simulation code for checking union properties
TBD

scottpurdy:
fixes #133


scottpurdy:
@subutai - fyi - the code is getting a bit messy. I can clean it up if we are going to share it or reuse it more but have been holding off until I know that this code is going to be used more in the future.

PR #142, opened 2015-03-05
TP MM plot method now calls inherited MM Base method

rmccall84:
Fixes https://github.com/numenta/nupic/issues/1825

Adapting Temporal Pooler Monitor Mixin in light of changes to its super class MonitorMixinBase


rmccall84:
@chetan51 this change relates to our other plot changes.

chetan51:
+1

PR #148, opened 2015-03-28
Initial commit handling orphan synapses

subutai:
Class implementing the General Temporal Memory algorithm with the added ability to weaken orphan segments. An orphan segment is one where it had some reasonably high input activity but the cell did not become active on the next time step.


subutai:
@chetan51 Per our discussion. This seems to handle the issue quite well. Let me know what you think. I still get occasional extra columns, but many fewer and it usually comes back down after a while.

subutai:
@chetan51 Don't merge anything though.

sensorimotor/sensorimotor/orphan_temporal_memory.py
```diff
@@ -0,0 +1,423 @@
...
+    orphanCells = set()
+
+    for cell in prevPredictiveCells:
+      column = self.columnForCell(cell)
+
+      if column in activeColumns:
+        activeCells.add(cell)
+        winnerCells.add(cell)
+        predictedColumns.add(column)
+      elif cell in prevMatchingCells:
```
chetan51: Is this necessary given https://github.com/numenta/nupic.research/pull/148/files#diff-1968c3cead7c67353b112847117974e5R253?

chetan51:
After a cursory glance, it looks good. I'm concerned about performance implications though, we should test that before merging.

sensorimotor/sensorimotor/orphan_temporal_memory.py
subutai: Good catch, that was left over from my first attempt. Thanks.

Issue #157, opened 2015-04-17
Union Pooler layer and first experiment


rmccall84:
Union Pooler will inherit from Spatial Pooler and will be partially modeled after Temporal Pooler.

Remaining tasks
- [x] Add reasonable param yamls
- [x] Output Union SDR trace to file
- [ ] Plot overlap between Union SDR with and without learning
- [ ] Verify correctness including vector operations (numpy.where())

rmccall84:
@scottpurdy fyi

PR #162, opened 2015-04-30
Union Temporal Pooler v0.1

rmccall84:
Fixes: https://github.com/numenta/nupic.research/issues/157

### DIFF
--- a/union_pooling/union_pooling/union_pooler.py
+++ b/union_pooling/union_pooling/union_pooler.py
@@ -1,0 +1,247 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have an agreement
+# with Numenta, Inc., for a separate license for this software code, the
+# following terms and conditions apply:
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import numpy
+
+from nupic.research.spatial_pooler import SpatialPooler
+
+
+
+REAL_DTYPE = numpy.float32
+
+
+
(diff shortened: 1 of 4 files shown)


rmccall84:
@chetan51 please review

union_pooling/union_pooling/union_pooler.py
```diff
@@ -0,0 +1,247 @@
...
+               dutyCyclePeriod=1000,
+               maxBoost=1.0,
+               seed=42,
+               spVerbosity=0,
+               wrapAround=True,
+
+               # union_pooler.py parameters
+               activeOverlapWeight=1.0,
+               predictedActiveOverlapWeight=10.0,
+               maxUnionActivity=0.20,
```
chetan51: Would `minUnionSparsity` be clearer?

union_pooling/union_pooling/union_pooler.py
```diff
@@ -0,0 +1,247 @@
...
+               synPredictedInc=0.5):
```
chetan51: Do we want this kind of learning in the first version? I thought we would only have spatial pooler learning, and do union pooling over the input.

union_pooling/union_pooling/union_pooler.py
```diff
@@ -0,0 +1,247 @@
...
+      self._bumpUpWeakColumns()
+      self._updateBoostFactors()
+      if self._isUpdateRound():
+        self._updateInhibitionRadius()
+        self._updateMinDutyCycles()
+
+    # Update and return the Union SDR
+
+    # Decrement pooling activation of all cells
+    self._poolingActivation -= 1
```
chetan51: Are we not doing the logistic function decay then?

union_pooling/union_pooling/union_pooler.py
rmccall84: Here's why I don't think so: this parameter serves as an upper bound on the ratio of active cells to total cells. So the min sparsity would vary e.g., from 0.0 to 0.2, so this name would suggest that min sparsity can be controlled. I also thought of using 'sparsity' in the name, but decided against it since I don't consider 20% sparse -- do you?

union_pooling/union_pooling/union_pooler.py
rmccall84: Thanks for pointing this out. I thought we might want to try this TP-style learning. (Perhaps this is something to discuss with others?) How about we should support this possibility but set the default value such that it has no effect?

union_pooling/union_pooling/union_pooler.py
rmccall84: I think it's a fine idea, but wasn't going to implement it for this initial version. If you think that logistic decay is critical let's discuss in person.

chetan51: So would this version build up density until it reaches a max, and then stay there?

PR #176, opened 2015-06-03
Union Pooler: Noise Robustness Experiment

rmccall84:
Noise Robustness Experiment

Data: Sequences generated from an alphabet.
Train Phase: Train network on sequences for some number of repetitions
Test phase: Input sequence pattern by pattern. Sequence-to-sequence
progression is randomly selected. At each step there is a chance that the
next pattern in the sequence is not shown. Specifically the following
perturbations may occur:
  1) random Jump to another sequence
  2) substitution of some other pattern for the normal expected pattern
  3) skipping expected pattern and presenting next pattern in sequence
  4) addition of some other pattern putting off expected pattern one time step
Goal: Characterize the noise robustness of the UnionPooler to various
perturbations. Explore trade-off between remaining stable to noise yet still
changing when sequence actually changes.


rmccall84:
@scottpurdy Let's discuss the design of the Noise Robustness Experiment in this PR soon

rmccall84:
@scottpurdy please review

rmccall84:
@scottpurdy I've made some efforts to improve the code based on our discussion the other day. review appreciated. ty

scottpurdy:
This looks a lot better. You could just have one parameter file and then modify the one or two parameters that you want to vary programmatically in a loop. That is probably easier than maintaining a whole bunch of files.

scottpurdy:
:+1:

rmccall84:
@scottpurdy @subutai Yes, I agree it is not ideal. Working on these experiments has underscored for me how a common testing utils could be useful for our research. One issue is making explicit all the parameters being used in the test. I went through the TM and the SP by hand to ensure this. I'm imaging a utility -- ParameterSetFactory --  that knows all of the parameters for different algorithms (SP, TM, etc) and has several sensible default parameter sets. Then users can request parameter sets based on these defaults, e.g., "give me a set of TM params varying cellsPerColumn from 8 to 32 in steps of 4." (Could also specify new experimental parameters). Armed with these parameter sets, you call a generic test harness like the one I've got in this PR (similar to one @chetan51 created) and you never have to mess around with param files...

cogmission:
May I suggest a look at Parameters.java in HTM.java?

https://github.com/numenta/htm.java/blob/master/src/main/java/org/numenta/nupic/Parameters.java

Along with some other useful features... The sheer enormity of the
parameters almost made this class write itself, in my experience... :-)



On Thu, Jun 18, 2015 at 12:43 PM, Ryan McCall <notifications@github.com>
wrote:

> @scottpurdy <https://github.com/scottpurdy> @subutai
> <https://github.com/subutai> Yes, I agree it is not ideal. Working on
> these experiments has underscored for me how a common testing utils could
> be useful for our research. One issue is making explicit all the parameters
> being used in the test. I went through the TM and the SP by hand to ensure
> this. I'm imaging a utility -- ParameterSetFactory -- that knows all of the
> parameters for different algorithms (SP, TM, etc) and has several sensible
> default parameter sets. Then users can request parameter sets based on
> these defaults, e.g., "give me a set of TM params varying cellsPerColumn
> from 8 to 32 in steps of 4." (Could also specify new experimental
> parameters). Armed with these parameter sets, you call a generic test
> harness like the one I've got in this PR (similar to one @chetan51
> <https://github.com/chetan51> created) an d you ne ver have to mess
> around with param files...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic.research/pull/176#issuecomment-113232491>
> .
>



-- 
*With kind regards,*

David Ray
Java Solutions Architect

*Cortical.io <http://cortical.io/>*
Sponsor of:  HTM.java <https://github.com/numenta/htm.java>

d.ray@cortical.io
http://cortical.io

scottpurdy:
The parameters we are using here may need to be tuned or whatever so we will need the values specified somewhere. A yaml file is fine for that. But if a specific experiment wants to vary specific parameters then it should load the single yaml file and then iterate through the parameter options it wants to change. It isn't clear to me how a utility factory would make that any easier.

cogmission:
Well first you can specify whole sets and subset relationships, you can
store and serialize known sets to make building easier, you can specify
defaults and check types to make sure they're the right type, you can check
ranges of known type so that a numeric type is within a known range; you
can settle on known conventional error messages when types are outside of
their range; you can do intersection and union operations on different sets
of parameters to unite known parameters (default) with other saved or
specified parameters; you can make sure that users always match a parameter
with its type - the known sets can correspond to the different algorithms
making it easier for users to specify a given subset of parameters.

need I continue because I can?  ;-)

On Thu, Jun 18, 2015 at 1:09 PM, Scott Purdy <notifications@github.com>
wrote:

> The parameters we are using here may need to be tuned or whatever so we
> will need the values specified somewhere. A yaml file is fine for that. But
> if a specific experiment wants to vary specific parameters then it should
> load the single yaml file and then iterate through the parameter options it
> wants to change. It isn't clear to me how a utility factory would make that
> any easier.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/numenta/nupic.research/pull/176#issuecomment-113242696>
> .
>



-- 
*With kind regards,*

David Ray
Java Solutions Architect

*Cortical.io <http://cortical.io/>*
Sponsor of:  HTM.java <https://github.com/numenta/htm.java>

d.ray@cortical.io
http://cortical.io

cogmission:
For HTM.java we have the ability to set ranges but haven't done it yet
because we don't have access to sanity data - something I would like to
eventually have the Parameters object do which would be nice. We need to be
able to specify sanity criteria and best practices criteria and be able to
enforce and report on them - would make life a **Lot** easier... 'specially
for folks like me that don't know one parameter from the next :-P

PR #180, opened 2015-06-11
Region for the Union Pooler and a demo for integrating it into the network

### LINKED ISSUE #69: Implement pooling region
Union pooler:
https://github.com/numenta/nupic.research/tree/master/union_pooling

This would be nice to support as a follow on:
https://github.com/numenta/nupic.research/blob/master/sensorimotor/sensorimotor/temporal_pooler.py

akhilaananthram:
The region for the union pooler encapsulates union_pooler.py and sets up the inputs for it.  It can be integrated into the network API.

network_integration_demo.py shows how this region can be integrated.  Because the current TPRegion does not output the active+predicted cells, I have a separate PR in nupic for that: https://github.com/numenta/nupic/pull/2206

@rmccall84 @scottpurdy 

Fixes #69


union_pooling/union_pooling/nupic_integration/custom_region/UPRegion.py
```diff
@@ -0,0 +1,349 @@
...
+      dataType="bool",
+      count=1,
+      constraints="bool"),
+  )
+
+  return unionSpec, otherSpec
+
+
+
+class UPRegion(PyRegion):
```
scottpurdy: `PoolingRegion` would be a better name in my opion, especially since we can probably use it for other pooling implementations

scottpurdy:
I think it would be better to put the pooling module at the same level as `union_pooler.py`. I'm not sure if the example is necessary. Using it in the experiments would probably serve as a better example but should be left for a separate PR.

akhilaananthram:
@scottpurdy I changed it to PoolingRegion and added a parameter to specify the type of region.

subutai:
> I'm not sure if the example is necessary.

I think it's helpful to have a small example showing how to use the region.  I agree that the experiment would serve as the ultimate test.

union_pooling/union_pooling/PoolingRegion.py
```diff
@@ -0,0 +1,367 @@
...
+    if t == "Byte":
+      return "multiple"
+    elif t == "bool":
+      return "bool"
+    else:
+      return ""
+
+  # Get arguments from pooler constructors, figure out types of
+  # variables and populate poolerSpec.
+  pArgTuples = _buildArgs(poolerClass)
```
scottpurdy: All of this logic is confusing. Do you understand it? Would it be simpler to just explicitly define `poolerSpec`?

union_pooling/union_pooling/PoolingRegion.py
```diff
@@ -0,0 +1,367 @@
...
+    """
+    if parameterName in self._poolerArgNames:
+      setattr(self._pooler, parameterName, parameterValue)
+    elif hasattr(self, parameterName):
+      setattr(self, parameterName, parameterValue)
+    else:
+      raise Exception("Unknown parameter: " + parameterName)
+
+
+  def __getstate__(self):
```
scottpurdy: No need for `__getstate__` or `__setstate__` if there isn't any logic to be done

union_pooling/union_pooling/PoolingRegion.py
akhilaananthram: If I explicitly define poolerSpec, then I would have to modify this section every time we make a change to UnionPooler's __init__ arguments.  Right now it gets all the arguments that the UnionPooler has by calling _buildArgs. So, if you make a change to UnionPooler, it will make that change here as well automatically.

akhilaananthram: We could move this section to PyRegion.py since this code is repeated in a lot of subclasses.

scottpurdy: Alright fine to leave as is. Not a bad idea to make this the default behavior but outside the scope of this PR

union_pooling/union_pooling/PoolingRegion.py
```diff
@@ -0,0 +1,352 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import numpy
+from nupic.bindings.math import GetNTAReal
+#from nupic.research.union_pooler import UnionPooler
```
scottpurdy: remove?

scottpurdy:
:+1: merge when you're ready

PR #182, opened 2015-06-15
changed the PoolerRegion to match the changes done to TPRegion

akhilaananthram:
Instead of the input being a list of active indices, it is a list of 0s and 1s, with 1s at the active indices.

I did this because I also changed TPRegion: https://github.com/numenta/nupic/pull/2206.

@scottpurdy @rmccall84


scottpurdy:
:+1:

PR #184, opened 2015-06-16
Modified Q-learning algorithm

chetan51:


chetan51:
cc: @subutai @scottpurdy @tomsilver

sensorimotor/sensorimotor/q_learner.py
```diff
@@ -71,11 +71,17 @@ def bestAction(self, state):
 
 
   def update(self, state, action, nextState, nextAction, reward):
+    targetValue = reward + self.gamma * self.value(nextState)
```
scottpurdy: I'd use parens to make the order of operations explicit

chetan51: Thanks, will do.

PR #187, opened 2015-06-17
Modified maze and sensing boundaries for better SDRs

chetan51:


chetan51:
cc: @subutai @scottpurdy @tomsilver

PR #197, opened 2015-06-18
T maze draft

tomsilver:
@chetan51 this seems to give pretty good results, it's a little wacky looking tho


chetan51:
Haha, it does look pretty wacky! This is a good first pass though.

However I'm still seeing consecutive encoding overlaps go up to 60% when approaching the objects (after making the turn). This is even with a 1Hz update rate.

screen shot 2015-06-18 at 11 16 27 pm

tomsilver:
Can you tell if the overlaps are due to the walls or the objects themselves?

tomsilver:
@chetan51 made some adjustments. is the code for generating the plot above in the repo somewhere?

chetan51:
Just tried the adjusted version, seeing the same results. It looks like the overlaps are due to the walls.

subutai:
Is this PR still valid? It needs to be updated to the new structure before we can merge

chetan51:
I don't think so, since it didn't actually improve the results. Closing.

PR #203, opened 2015-06-25
reverted to original q learning update rule

tomsilver:
This performs about the same as the current update rule for the T maze (which is not very well), but the update rule is cleaner. We may revisit and try to understand better why the algorithm mistakenly learns a high Q value for turning right into a negative reward.

@chetan51

### DIFF
--- a/sensorimotor/sensorimotor/q_learner.py
+++ b/sensorimotor/sensorimotor/q_learner.py
@@ -31,7 +31,7 @@ from sensorimotor.reinforcement_learner import ReinforcementLearner
 class QLearner(ReinforcementLearner):
 
   def __init__(self, actions,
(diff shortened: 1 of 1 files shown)


sensorimotor/sensorimotor/q_learner.py
```diff
@@ -31,7 +31,7 @@
 class QLearner(ReinforcementLearner):
 
   def __init__(self, actions,
-               alpha=0.2, gamma=0.8, elambda=0.3,
+               alpha=0.9, gamma=0.8, elambda=0.3,
```
chetan51: Is this parameter change intentional?

tomsilver:
@chetan51 changed back alpha

PR #206, opened 2015-06-29
another simple demo of encoder check

tomsilver:
shows that higher resolution performs significantly better for rsde


tomsilver:
@scottpurdy please review no rush

encoder_quality/rdse_check_demo.py
```diff
@@ -0,0 +1,59 @@
...
+                                           w=23, n=500, offset=0.0)
+
+
+  distance_function = lambda x,y : abs(x-y)
+
+  sample_generator = lambda : np.random.uniform(minval, maxval)
+  input_pairs_source = encoder_check.InputTripleCreator(sample_generator)
+
+  err1 = encoder_check.encoderCheck(encoder1, distance_function,
+  	                               input_pairs_source)
```
scottpurdy: nit: indentation here and pretty much every other place there is a line break

scottpurdy:
You might want to put a module-level doc string explaining what this is showing. I would be careful about stating that the higher resolution performs better in an absolute sense because for some applications it wouldn't be. I keep saying this and am getting the feeling that you disagree so maybe we should discuss it. But you could say that for this distance function, a finer-grained resolution produces better SDRs.

tomsilver:
@scottpurdy addressed your feedback

PR #214, opened 2015-07-06
Refactor agents for improved experiment framework

chetan51:
Now we just have to run the agent (no need to run the proxy server), and then run the simulation.


chetan51:
cc: @tomsilver

PR #217, opened 2015-07-07
More informative overlap plots

chetan51:


chetan51:
cc: @subutai @tomsilver

PR #218, opened 2015-07-08
Made overlap plots optional and showing more information in them

chetan51:


chetan51:
cc: @subutai @tomsilver

PR #220, opened 2015-07-08
Show reset lines in plots

chetan51:
So you can see where sequences start and end.


chetan51:
cc: @subutai @tomsilver

PR #221, opened 2015-07-08
Tuned encoder params for better overlaps

chetan51:
Overlaps before:

before-tuning-radius-7

Overlaps after:

after-tuning-radius-10-timestep-1 5

(x and y axes are timesteps)


chetan51:
cc: @subutai

PR #244, opened 2015-07-27
Multi-category classification

BoltzmannBrain:


classification/classification_network.py
```diff
@@ -75,10 +73,17 @@
     "pamLength": 3,
 }
 
+
+# TODO: replace with SEQ_CLASSIFIER_PARAMS
 CLA_CLASSIFIER_PARAMS = {"steps": "1",
                          "implementation": "py"}
+KNN_CLASSIFIER_PARAMS = {
```
BoltzmannBrain: I added this in b/c I like using it to run validation experiments; the kNN classifier is very intuitive as to what the exact classification probabilities should be, and the `compute` method works.

classification/classify_active_cells.py
```diff
@@ -146,14 +139,11 @@ def run(net, numRecords, partitions, outFile):
...
-      # TODO[continued] but it is not implemented in the tm_py, so we use this work around for now.
+      # but it is not implemented in the tm_py, so we use this work around for now.
       # predictiveCells = tmInstance.getOutput("predictedActiveCells")
       predictiveCells = tmInstance.predictiveCells
-      predictedActiveCells = numpy.intersect1d(activeCells, predictiveCells)
-      if len(predictiveCells) >0: #TODO: this line to be removed. for debugging purposes.
-        #print "predictiveActiveCells: %s" %predictedActiveCells
-        pass
-      
+      predictedActiveCells = [cell for cell in predictiveCells if activeCells[cell]==1]
```
BoltzmannBrain: works for numpy arrays and lists

BoltzmannBrain:
@marionleborgne please review. The changes here that would affect your experiments is the addition of the `numCategories` parameter when creating a sequence classification network instance; this is what we discussed yesterday.

classification/classification_network.py
marionleborgne: :+1: cool

marionleborgne:
waiting on nupic #2339 to be able to merge

classification/classification_network.py
```diff
@@ -140,16 +131,29 @@ def createSensorRegion(network, sensorType, encoders, dataSource):
...
   @param dataSource   (RecordStream)  Sensor region reads data from here.
 
+  @param numCats      (int)           Number of possible categories per record.
+
   @return             (Region)        Sensor region of the network.
   """
   # Sensor region may be non-standard, so add custom region class to the network
-  if sensorType.split(".")[1] not in PY_REGIONS:
+  SENSOR_NAME = sensorType.split(".")[1]
+  SENSOR_MODULE = SENSOR_NAME  # conveniently have the same name
```
marionleborgne: It's a minor detail, but is there a specific reason why you decided to go with upper case names for SENSOR_NAME and SENSOR_MODULE? this is usually used for global module variables or constants. Just curious.

marionleborgne:
@BoltzmannBrain Thanks for the changes. I like that you added logic to import the correct sensor based on its type.

classification/classification_network.py
BoltzmannBrain: Hmm I can't think of why I'd do that... thanks for catching it!

BoltzmannBrain:
@marionleborgne should be good to go

marionleborgne:
:+1: thanks!

PR #277, opened 2015-08-19
RES-37 integrate union pooler

marionleborgne:
@BoltzmannBrain @akhilaananthram 
It's still a WIP. I added the notion of `networkConfigurations` to be able to run different network for the classification experiments of NLP/sensor data. The idea is to be able to create, link, validate networks in various setups (e.g. `sensor -> SP -> TM -> CLAClassifier `,  `sensor -> SP -> TM -> KNNClassifier `,  `sensor -> SP -> TM -> UP -> CLAClassifier `,  `sensor -> SP -> KNNClassifier `etc... )


marionleborgne:
@BoltzmannBrain ready for review

marionleborgne:
@BoltzmannBrain ready for review. I included the changes we discussed today.

classification/classification_network.py
```diff
@@ -21,65 +21,41 @@
 # ----------------------------------------------------------------------
 """
 The methods here are a factory to create a classification network:
-  encoder -> SP -> TM -> (UP) -> classifier
+  encoder -> (SP) -> (TM) -> (UP) -> classifier
```
BoltzmannBrain: If we can specify the regions we should say so: 

    """
    The methods here are a factory to create a classification network of any of sensor, SP, TM, UP, and classifier regions.
    """

classification/classification_network.py
```diff
@@ -99,63 +75,61 @@ def createEncoder(newEncoders):
...
+                        sensorType,
+                        sensorParams,
+                        dataSource,
+                        encoders):
   """
   Initializes the sensor region with an encoder and data source.
 
   @param network      (Network)
+  
+  @param regionName   (str) Name for this region
```
BoltzmannBrain: Cleanup docstring format. However you want to style it is fine, just be consistent. Please make these changes throughout.

classification/classification_network.py
```diff
@@ -213,28 +189,26 @@ def createTemporalMemoryRegion(network):
 
 
 
-def createClassifierRegion(network, classifierType, classifierParams):
+def _createClassifierRegion(network,
```
BoltzmannBrain: Why can't these params be `network, regionName, classifierParams`, and then `classifierType` is just within `classifierParams`? Then all the `_createXRegions()` methods take the same args, and if we want we can make a generic `addRegions()` method.

classification/classification_network.py
```diff
@@ -99,63 +75,61 @@ def createEncoder(newEncoders):
...
-  encoder.addMultipleEncoders(newEncoders)
+  encoder.addMultipleEncoders(encoders)
   return encoder
 
 
 
-def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
+def _createSensorRegion(network,
+                        regionName,
+                        sensorType,
```
BoltzmannBrain: Can `sensorType` be w/in `sensorParams`?

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
-                                      classifierParams)
+def _enableRegionLearning(network,
+                          trainedRegionNames,
+                          regionName,
+                          recordNumber):
+  """
+  Enable learning for a specific region.
+  
+  @param network (Network) the network instance 
+  @param trainedRegionNames (list) regions that have been trained on the
```
BoltzmannBrain: what's this used for?

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
 
+
+def _stopLearning(network, trainedRegionNames, recordNumber):
+  """
+  Disable learning for all trained regions.
+  
+  @param network (Network) the network instance 
+  @param trainedRegionNames (list) regions that have been trained on the 
+  input data.
+  @param recordNumber (int) value of the current record number
```
BoltzmannBrain: what's this used for?

marionleborgne: for the test phase, you need to disable learning

classification/classification_network.py
marionleborgne: you need to enable learning of some regions during training

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
+  #   RecordSensor region allows us to specify a file record stream as the
+  #   input source via the dataSource attribute.
+  dataSource = FileRecordStream(streamID=inputFile)
+
+  network = createNetwork(dataSource,
+                          encoders,
+                          networkConfiguration)
+
+  # Need to init the network before it can run.
+  network.initialize()
```
BoltzmannBrain: Good to have this here before returning the network instance b/c this checks the links and region widths :+1:

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
   return network
+
+
+
+def _findNumberOfPartitions(networkConfiguration, numRecords):
```
BoltzmannBrain: This method should be cleaner, e.g.

    partitionNames = []
    regions = (SP_REGION_NAME, TM_REGION_NAME, UP_REGION_NAME)
    for r in regions:
        if networkConfiguration[r]["enabled"]:
            partitionNames.append(r)

    partitionNames.append(CLASSIFIER_REGION_NAME)
    partitionNames.append(TEST_PARTITION_NAME)

    numPartitions = len(partitionNames)

    partitionIndices = [numRecords * i / numPartitions for i in xrange(numPartitions)]

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
+  partitionIndices = [numRecords * i / numPartitions
+                      for i in range(0, numPartitions)]
+
+  return zip(partitionNames, partitionIndices)
+
+
+
+def _setScalarEncoderMinMax(networkConfiguration, minval, maxval):
+  """
+  Introspect an input file and set the min and max values.
```
BoltzmannBrain: Introspect? What input file?

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
+  scalarEncoderParams = (networkConfiguration
+                         [SENSOR_REGION_NAME]
+                         ["encoders"]
+                         [SCALAR_ENCODER_NAME])
+  scalarEncoderParams["minval"] = minval
+  scalarEncoderParams["maxval"] = maxval
+
+
+
+def _findCSVMinMax(fileName):
```
BoltzmannBrain: This is implemented for nupic record streams (see below). Did you try already?
- [FileRecordStream.getStats()](https://github.com/numenta/nupic/blob/master/nupic/data/file_record_stream.py#L494)
- [RecordStream.getFieldMin()](https://github.com/numenta/nupic/blob/master/nupic/data/record_stream.py#L365) and [getFieldMax()](https://github.com/numenta/nupic/blob/master/nupic/data/record_stream.py#L382)

classification/generate_sensor_data.py
```diff
@@ -27,27 +27,31 @@
...
-                 signal_mean=SIGNAL_MEAN,
-                 signal_period=SIGNAL_PERIOD,
-                 number_of_points=NUM_RECORDS,
-                 signal_amplitude=SIGNAL_AMPLITUDE,
-                 noise_amplitude=DEFAULT_WHITE_NOISE_AMPLITUDE):
-  fileName = "white_noise_%s" % noise_amplitude
-
-  if not dataDir:
-    dataDir = DATA_DIR
+def generateData(dataDir,
```
BoltzmannBrain: You use `open()` and `close()` for the file here. You should avoid that logic at all costs, and use the `with open()` context manager, **always**.

classification/generate_sensor_data.py
```diff
@@ -65,31 +69,27 @@ def generateData(dataDir=None,
...
   fileHandle.close()
 
-  print "Data generated. File saved to %s/%s.csv" % (dataDir, fileName)
-
-
-
-if __name__ == "__main__":
-  for whiteNoiseAmplitude in WHITE_NOISE_AMPLITUDES:
-    generateData(noise_amplitude=whiteNoiseAmplitude)
+  outFilePath = os.path.join(dataDir,
```
BoltzmannBrain: this is already saved in `filename`, so why repeat this operation?

classification/network_params.py
```diff
@@ -0,0 +1,148 @@
...
+    },
+  UP_REGION_NAME:
+    {
+      "enabled": False,
+      "params": UP_PARAMS,
+    },
+  CLASSIFIER_REGION_NAME:
+    {
+      "type": CLA_CLASSIFIER_TYPE,
+      "params": CLA_CLASSIFIER_PARAMS
```
BoltzmannBrain: add comma

classification/network_params.py
```diff
@@ -0,0 +1,148 @@
...
+
+# Union pooler region params
+# TODO: Don't know what the UP params are yet.
+UP_PARAMS = {}
+
+# A network configuration specifies what regions to add to a network and what
+# parameters to use for each one of these regions.
+# Classifier and Sensor regions are always enabled but can have different 
+# types. Make sure the params matches the type
+NETWORK_CONFIGURATION = {
```
BoltzmannBrain: Didn't we agree these network configs would be in individual json files?

classification/settings.py
```diff
@@ -21,21 +21,25 @@
 # ----------------------------------------------------------------------
 
 
-SIGNAL_TYPES = ["white_noise"]
-RESULTS_DIR = "results"
-DATA_DIR = "data" 
-MODEL_PARAMS_DIR = 'model_params'
+# Parameters to generate the artificial sensor data
+OUTFILE_NAME = "white_noise"
```
BoltzmannBrain: would be helpful to rename this file to white_noise_settings.py

classification/test_sensor_data_classification.py
```diff
@@ -39,35 +92,114 @@ class TestSensorDataClassification(unittest.TestCase):
...
+                                signalPeriod,
+                                noiseAmplitude,
+                                sensorType.split(".")[1],
+                                spEnabled,
+                                tmEnabled,
+                                upEnabled,
+                                classifierType.split(".")[1])
+              print expParams
+
+              inputFile = generateData(DATA_DIR,
```
BoltzmannBrain: Better to store this as `self.inputFile` so the `tearDown()` can be more explicit -- i.e. not guess what the file name is. This way if the file naming scheme in `generateData()` changes you won't  have to remember to change it here too.

classification/test_sensor_data_classification.py
```diff
@@ -39,35 +92,114 @@ class TestSensorDataClassification(unittest.TestCase):
...
+              network = configureNetwork(inputFile,
+                                         networkConfig)
+
+              (numCorrect,
+               numTestRecords,
+               predictionAccuracy) = runNetwork(network,
+                                                networkConfig,
+                                                NUM_RECORDS)
+
+              if (noiseAmplitude == 0
```
BoltzmannBrain: Would be much cleaner to have separate test methods for each of these. At the very least you should include fail messages in the assertions so we can see clearly what fails and why.

classification/classification_network.py
BoltzmannBrain: And didn't we decide yesterday not to partition the training sets like this?

classification/network_params.py
```diff
@@ -0,0 +1,148 @@
...
+CLASSIFIER_REGION_NAME = "classifierRegion"
+
+# Sensor types
+SENSOR_TYPE = "py.RecordSensor"
+
+# Encoder names
+SCALAR_ENCODER_NAME = "scalarEncoder"
+SCALAR_ENCODER_FIELD_NAME = "y"
+
+# Classifier types
```
BoltzmannBrain: This is confusing... why not have a `CLASSIFIER_TYPE` the user defines as either "py.CLAClassifierRegion" or "py.KNNClassifierRegion"?

classification/classification_network.py
```diff
@@ -246,109 +220,368 @@ def createClassifierRegion(network, classifierType, classifierParams):
...
+        trainedRegionNames.append(partitionName)
+        _enableRegionLearning(network,
+                              trainedRegionNames,
+                              partitionName,
+                              recordNumber)
+
+    # Evaluate the predictions on the test set.
+    if recordNumber >= testIndex:
+      if classifierRegion.type == KNN_CLASSIFIER_TYPE:
+        inferredValue = classifierRegion.getOutputData("categoriesOut").argmax()
```
BoltzmannBrain: `argmax()` is not a good way to get the inference value b/c it handles ties by always taking the last occurrence. Try this:

    # The use of numpy.lexsort() here is to first sort by labelFreq, then sort
    # by random values; this breaks ties in a random manner.
    inferenceValues = classifierRegion.getOutputData("categoriesOut")
    randomValues = numpy.random.random(inferenceValues.size)
    inferredValue = numpy.lexsort((randomValues, inferenceValues))[:-1]

classification/classification_network.py
BoltzmannBrain: I mean specifically the line I commented on... it seems like `trainedRegionNames` is useless, no?

classification/classification_network.py
BoltzmannBrain: as above, specifically `recordNumber`...

BoltzmannBrain:
@marionleborgne done with this round of review.

classification/classification_network.py
marionleborgne: good point. i forgot to update this docstring

classification/classification_network.py
marionleborgne: Thanks, I didn't know about it. That works.

classification/classification_network.py
marionleborgne: good point. changed that to what you were proposing

classification/classification_network.py
marionleborgne: yes. made the change. thanks for pointing this out

classification/classification_network.py
marionleborgne: well i use it for printing useful phase info. it's true that functionally it doesn't change anything. But i find this info useful to know what is going on in the network and how it is learning (i.e. what are the regions learning for each phase) I'd rather keep it, at least in nupic.research

classification/classification_network.py
marionleborgne: as explained above, I use this for printing out useful phase info. It' helpful to validate what region is learning / stops learning, at what point in the training set

classification/classification_network.py
marionleborgne: correct. I made a change so that we are sure to test and train on the same length of data, regardless of the network configuration.

classification/generate_sensor_data.py
marionleborgne: updated. thanks

classification/generate_sensor_data.py
marionleborgne: updated

classification/settings.py
marionleborgne: agreed. renamed to "sensor_data_exp_settings.py" since i'll be using more than white noise for the experimments in the future.

classification/network_params.py
marionleborgne: we agreed that all network config info + sensor params should be in the same settings file. I think it's convenient to leave it this way for now. I made sure experiments settings and network settings are separate. Is that ok?

classification/test_sensor_data_classification.py
marionleborgne: good point. I included that.

classification/classification_network.py
marionleborgne: thanks for the tip

classification/test_sensor_data_classification.py
marionleborgne: I agree that this should ultimately be modified. It's a (temporary) convenient way for me to loop though experiments params. Ultimately I'll move to the exp suite that chetan demo'ed, it's a better way to run and validate my epxeriements. I'll to keep the code this way, until I move to that exp suite for the sensor data experiments - one of my next steps. I'll add unit testing after all the experiments are done running via the expsuite framework.

marionleborgne:
@BoltzmannBrain done with updates

classification/network_params.py
BoltzmannBrain: It may be convenient for now, but this is core functionality that justifies having a network factory in the first place. We should implement the json configuration --> network instance pipeline now.

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
+  @param networkConfiguration: (dict) the configuration of this network.
+  @param dataSource: (RecordStream) the input source
+  """
+  scalarEncoderParams = (networkConfiguration
+                         [SENSOR_REGION_NAME]
+                         ["encoders"]
+                         [SCALAR_ENCODER_NAME])
+  fieldName = scalarEncoderParams["fieldname"]
+  scalarEncoderParams["minval"] = dataSource.getFieldMin(fieldName)
+  scalarEncoderParams["maxval"] = dataSource.getFieldMax(fieldName)
```
BoltzmannBrain: no newline at the end of the file

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
+                                  numCorrect,
+                                  numTestRecords))
+  print results
+
+  return numCorrect, numTestRecords, predictionAccuracy
+
+
+
+def configureNetwork(inputFile,
+                     networkConfiguration):
```
BoltzmannBrain: keep the params on one line when you can

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
+  sensorRegion = _createSensorRegion(network,
+                                     sensorParams,
+                                     dataSource)
+
+  # Keep track of the previous region name and width to validate and link the 
+  # input/output width of two consecutive regions.
+  previousRegion = SENSOR_REGION_NAME
+  previousRegionWidth = sensorRegion.encoder.getWidth()
+
+  # Create SP region, if enabled.
```
BoltzmannBrain: These if blocks should be generalized into one `createRegion(regionParams)` method.

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
-                                      classifierType,
-                                      classifierParams)
+  return sorted(partitions.items(), key=lambda x: x[1])
 
-  validateRegions(sensor, sp, tm, classifier)
 
-  linkRegions(network)
 
-  return network
+
```
BoltzmannBrain: Why are there so many blank lines??

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
+def _enableRegionLearning(network,
+                          trainedRegionNames,
+                          regionName,
+                          recordNumber):
+  """
+  Enable learning for a specific region.
+  
+  @param network: (Network) the network instance 
+  @param trainedRegionNames: (list) regions that have been trained on the 
+  input data.
```
BoltzmannBrain: indent line continuation

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
+  @param regionName: (str) name of the current region  
+  @param recordNumber: (int) value of the current record number 
+  """
 
-    # TODO: should we check if TM output width matches classifier input width? 
+  network.regions[regionName].setParameter("learningMode", True)
+  phaseInfo = ("-> Training '%s'. RecordNumber=%s. Learning is ON for %s, "
+               "but OFF for the remaining regions." % (regionName,
+                                                       recordNumber,
+                                                       trainedRegionNames))
```
BoltzmannBrain: Why? Can't I train a region and then turn off its learning?

classification/test_sensor_data_classification.py
```diff
@@ -39,45 +92,127 @@ class TestSensorDataClassification(unittest.TestCase):
   Test classification results for sensor data.
```
BoltzmannBrain: Single line docstrings should be on a single line... `"""Test classification results for sensor data."""`

classification/classification_network.py
```diff
@@ -166,189 +115,356 @@ def createSensorRegion(network, sensorType, encoders, dataSource, numCats):
...
+    # TODO: not sure about the UP region width params. This needs to be updated.
+    _validateRegionWidths(previousRegionWidth,
+                          upRegion.getSelf().cellsPerColumn)
+    _linkRegions(network,
+                 SENSOR_REGION_NAME,
+                 previousRegion,
+                 UP_REGION_NAME)
+    previousRegion = UP_REGION_NAME
+
+  # Create classifier region (always enabled)
```
BoltzmannBrain: Is creating and linking the classifier region necessary to hardcode?

BoltzmannBrain:
@marionleborgne done with this round of review.

The main purpose of the network factory is to create network instances for a variety of configurations, so it needs to be generalizable. Currently there's too much specific code:
- We should not include `_findNumberOfPartitions()`. Whatever experiment runner or model is running the network should just specify the partitions as an argument to runNetwork() -- i.e. the logic for partitioning training/testing data is out of scope for the network factory.
- Creating/linking a classifier region, or any region for that matter, should not be mandatory. We would like to add this factory to NuPIC where there are many non-classification use cases.
- Simply passing a json of params should return the user with a runnable network instance. The an experiment runner could simply pass in "region_params.json" and "network_config.json". For example the former could be lines 47-120 of your "classification/network_params.py" and the latter could be lines 121-157.

Also there's still a bunch of formatting issues. Please run pylint before pushing (and use the nupic rc file).

classification/network_params.py
marionleborgne: replaced by json file

marionleborgne:
@BoltzmannBrain 
* `_findNumberOfPartitions()` is now out of `classification_network.py` and the `partions` arg is passed when running the network.
* the network config is now a json

marionleborgne:
@BoltzmannBrain regarding the `network_config.json` config file, I think I should validate and map the value of each "xRegionConfig" key (which is a dict describing the region config) to a `RegionConfig` class. This `regionConfig` class would have the following attributes:
* `regionName` (str) name of the region in the network
* `regionEnabled` (bool) 1 if region enabled in the network. 0 otherwise.
* `regionType` (str) Type of the region. E.g. "py.SPRegion"
* `regionParams` (dict) nupic parameters for this region
* `regionConfigId`  (str) ID of this region config. E.g. "sensorRegionConfig". This value corresponds to the key of that regionConfig in the `network_config.json`.

That way I can validate the structure of the JSON config file when instantiating each `RegionConfig` class. And also: all the setters and getters currently in the network factory `classification_network.py` will go in this `RegionConfig` class.  don't think it makes sense to have them in the network factory like what we discussed yesterday.

I think it will be easier to understand, use,  and validate the format of the `network_config.json` this way.

Does it make sense?

(PS: As I mentioned, I've to stop woking on this PR for a bit, but I'm happy to help with that later on)

PR #288, opened 2015-09-09
Generalize sequence classification factory for many network configurations

BoltzmannBrain:
Wraps in #277


BoltzmannBrain:
@marionleborgne please review. A few notes on my changes:
- Specifying partitions via namedtuples -- assigns names, as well as the numerical index, to each member. I left the factory `runNetwork()` method agnostic as to the list items' field types (list or tuple) because then a user can still have mutable partitions if they want.
- I got rid of the getters and setters; my rule of thumb is until you need anything more than a simple attribute access, don't write getters and setters. I left `getEncoderParam()` b/c it's good for returning `None` when an encoder param doesn't exist.
- Added logging facility.

 I think it's in good shape now for us to move on. As we discussed earlier today, your proposed `RegionConfig` class will be implemented in a followup PR.

Also, the sensor data test is okay, but I still think it doesn't serve to validate the sequence classification network model. For validation you would need to test against a known sequence, asserting both the correct and incorrect expected classifications.

marionleborgne:
@BoltzmannBrain: Thanks, Alex.  I'm done with the review.  Looks good to me. I just left a small comment.

Also, this might have been a bit unnecessary to create a separate PR from #277. I'm seeing mostly code cleanup in your commits, so I am not sure how the network factory became more generalized. Is the plan to include that here or will this be in another PR? Seems like that might be the latter based on your previous comment about the `NetworkConfig` class. Either way is fine with me, just wanted to clarify. 

If you don't want to add additional functionalities as part of this PR, then :+1. Merge away :smile:  Let me know if I can close #227 too.

BoltzmannBrain:
@marionleborgne I responded to your comment and made some add'l changes to generalize the dynamic import logic so we can import more than sensor regions (i.e. UP region). Please take a look and let me know if I can merge.

This PR wraps in #277, hence all of your commits. So merging this PR will merge all your commits, my commits, and auto close #277.

marionleborgne:
@BoltzmannBrain The unittests don't pass after the commit "split classification_network methods (...)". 

 Just checked and they did pass at the previous commit. Can you make sure the test pass after your changes?

BoltzmannBrain:
@marionleborgne what unittests? All I see is test_sensor_data_classification.py, and this passes.

PR #294, opened 2015-09-15
Update LSTM with scalar encoder

ywcui1990:
@subutai 
Summary

* Add separate scalar encoders for input and output layer of the LSTM network
* Parameter tuning of LSTM network for better performance
* Readout most likely point estimation from the scalar LSTM network, report NRMSE on that
* Use fixed random seed for LSTM to get reproducible result

Current performance of the LSTM-scalar Encoder network
negLL Train 1.789784 Test 1.946635
NRMSE, Train 0.405364, Test 0.504483

image

The current best performance of LSTM-scalarEncoder network was achieved by:
1. using the single stream normalized scalar data (not the encoded data) as input to the network
2. using a relatively small number of LSTM cells (n=20). Larger numbers are prune to overfitting
3. using Sigmoid as the output layer. Softmax gives slightly worse result
4. requires 60 epochs, more epochs may not necessarily lead to better performance, probably due to over fitting.


floybix:
Since you mentioned RMSE, just wanted to make sure you know about MASE which is appropriate for forecasts: https://en.wikipedia.org/wiki/Mean_absolute_scaled_error

subutai:
@floybix Thanks. Yes, MAPE (which is a variant of MASE) is actually the default metric in NuPIC. We use RMSE here since mean squared error is what is optimized in LSTMs.

ywcui1990:
Thanks Felix. The default error metric in nupic is "Alternative" Mean
Absolute Percent Error, which is very similar to MASE.
https://github.com/numenta/nupic/blob/master/nupic/frameworks/opf/metrics.py#L612

We used NRMSE here mainly because some of the early LSTM studies used the
that metric.

floybix:
Ok, got it. Cheers.

PR #305, opened 2015-09-18
add nlp code to new research repo

BoltzmannBrain:


subutai:
:+1: The README needs to be updated, but we can do that later.

BoltzmannBrain:
Yes I have an almost finished README with examples, but I've been having issues importing in htmresearch. I posted a [stack overflow question here](http://stackoverflow.com/questions/32644387/python-importerror-no-module-named-but-the-module-exists).

subutai:
Are you sure you have cleared out your previous fluent installs?

PR #313, opened 2015-09-23
Update PoolingRegion to work with Network API (specifically in NLP research)

chetan51:


chetan51:
@BoltzmannBrain @subutai Please review that these changes make sense.

htmresearch/regions/PoolingRegion.py
```diff
@@ -243,7 +249,14 @@ def compute(self, inputs, outputs):
     The guts of the compute are contained in the self._poolerClass compute() call
     """
     activeCells = inputs["activeCells"]
-    predictedActiveCells = inputs["predictedActiveCells"]
+    predictedActiveCells = inputs["predictedActiveCells"] if (
```
subutai: If this is a required input, should we throw an exception if not connected? If so, the exception can be thrown in initialize, which is always called once before the first all to compute().

htmresearch/regions/PoolingRegion.py
```diff
@@ -243,7 +249,14 @@ def compute(self, inputs, outputs):
...
+      "predictedActiveCells" in inputs) else numpy.zeros(self._inputWidth)
+
+    resetSignal = False
+    if 'resetIn' in inputs:
+      assert len(inputs['resetIn']) == 1
```
subutai: This check can be in initialize() and should throw an exception not an assert.

subutai:
:+1:

htmresearch/regions/PoolingRegion.py
chetan51: This is not a required input. In fact in the NLP experiment that uses pooling, it's connected to a sensor directly, which wouldn't produce predicted active cells.

PR #326, opened 2015-10-12
Experiments with Data Aggregation Window

ywcui1990:
Experimenting an algorithm that automatically select data aggregation window with wavelet transformations


rcrowder:
:+1: Is this the first use of wavelets in nupic?

ywcui1990:
Probably the first use of wavelets. It's not ready to be a incorporated into nupic at this point though. We tried to use wavelets to heuristically pick a reasonable data aggregation window. These are some very preliminary experiments.

subutai:
FWIW we used Gabor filters for vision applications in the previous generation of the HTM algorithms.

rcrowder:
Thanks for the update guys. I've fond memories of wavelet research from my personal research back in early 90's into global illumination (black-body radiation/radiosity simulation, and encoding bi-directional reflection distribution functions). Keen to see them more exploited, and at some point revisit them for audio work.

PR #354, opened 2015-11-06
scripts to run the param finder, and to freeze it with cx_freeze.

marionleborgne:
cc @ywcui1990


htmresearch/frameworks/utils/param_finder.py
```diff
@@ -0,0 +1,380 @@
...
+  output = np.zeros([len(widths), len(data)])
+  for ind, width in enumerate(widths):
+    wavelet_data = wavelet(min(10 * width, len(data)), width)
+    output[ind, :] = _convolve(data, wavelet_data,
+                               mode='same')
+  return output
+
+
+
+def read_csv_files(fileName):
```
ywcui1990: This function (and the csv module) is only needed for loading the example data. We may not need to include it in the param_finder

requirements.txt
```diff
@@ -2,3 +2,4 @@
 pandas
 plotly
```
ywcui1990: why is plotly required?

ywcui1990:
:+1:

requirements.txt
marionleborgne: I don't know, it was there before

htmresearch/frameworks/utils/param_finder.py
marionleborgne: We may or may not. The param finder will need to read data from the file system in Unicorn. Also, the actual script in unicorn will most likely be structured differently / moved somewhere else, so it's probably ok to leave this script like this for now. What are your thoughts?

PR #357, opened 2015-11-09
[fix] removed cx_freeze dependency

marionleborgne:
removed cx_freeze dependency since it is not required by htmresearch package.


lscheinkman:
:+1: Thank you

Issue #333, opened 2015-10-23
Specify SP params via TP region config


BoltzmannBrain:
When creating a network with a TP region in the sequence classification factory, you can only specify the TP params and not the underlying SP params. That is, the TP subclasses the SP, so the factory assumes [these standard SP params](https://github.com/numenta/nupic/blob/master/src/nupic/research/spatial_pooler.py#L113-130). Trying to specify SP params in the TP region config json will raise an error because they are not defined in the TP region's spec.

subutai:
There's a Pythonic trick to do this easily and automatically. You can simply introspect the appropriate classes and add params automatically to the spec.  This is used in some of the other regions.

PR #364, opened 2015-11-24
Simple sequence classification

ywcui1990:
* Test sequence classification network with simple dataset, the goal is to classify two category of sequences, animal-eats-vegetable or vegetable-eats-animal
* Words are encoded with 4K cortical.io sdr
* TM learns words transitions
* TPregion forms a pooled representation based on active/predicted-active cells in TM for each sentence
* Simple KNN classifier is used to classify sentences

@BoltzmannBrain I made some changes to TemporalPoolerRegion so that I can set SpatialPooler related parameters in the network config files. Could you review the changes in TemporalPoolerRegion.py? Thanks!


htmresearch/regions/TemporalPoolerRegion.py
```diff
@@ -44,6 +44,10 @@ def _getPoolerClass(name):
         " union, unionMonitored" % (name))
 
 
+def _getParentSpatialPoolerClass():
+    return UnionTemporalPooler.__bases__[0]
```
BoltzmannBrain: Would be better if you called `_getPoolerClass()` so this isn't hard coded for the union implementation, and return all the bases (not just the first in the list).

htmresearch/regions/TemporalPoolerRegion.py
```diff
@@ -139,6 +143,19 @@ def getConstraints(arg):
       constraints=getConstraints(argTuple[2]))
     poolerSpec[argTuple[0]] = d
 
+  # Get arguments from spatial pooler constructors, figure out types of
+  # variables and populate poolerSpec.
+  # This allows setting SP parameters
```
BoltzmannBrain: Does this fix #333?

htmresearch/regions/TemporalPoolerRegion.py
```diff
@@ -223,9 +242,9 @@ def __init__(self, columnCount, inputWidth, poolerType, **kwargs):
     # These calls whittle down kwargs and create instance variables of TemporalPoolerRegion
     self._poolerClass = _getPoolerClass(poolerType)
     pArgTuples = _buildArgs(self._poolerClass, self, kwargs)
-
+    pArgTuplesSP = _buildArgs(_getParentSpatialPoolerClass(), self, kwargs)
```
BoltzmannBrain: Following from earlier comment (line 48), here is where you would specify the first parent `_getParentSpatialPoolerClass()[0]`.

BoltzmannBrain:
@ywcui1990 I left some comments on TemporalPoolerRegion.py. Taking a look at projects/nlp/run_simple_classification_experiment.py, it's difficult to follow. Please use Python conventions for naming variables and functions, and PEP8 and our style conventions.

ywcui1990:
@chetan51

htmresearch/regions/TemporalPoolerRegion.py
ywcui1990: Yes, I think it fix that issue

htmresearch/regions/TemporalPoolerRegion.py
ywcui1990: I changed _getParentSpatialPoolerClass to be more general. It will always return the spatial pooler parent of the poolerClass in TPRegion

subutai:
> Taking a look at projects/nlp/run_simple_classification_experiment.py, it's difficult to follow.

@BoltzmannBrain Valid point. Assuming the TemporalPoolerRegion.py changes are ok, I think we should merge with the script as-is (given the changes already made).

BoltzmannBrain:
:+1:

PR #372, opened 2015-12-08
Fixes Initialization Problem of Temporal Pooler Region

ywcui1990:
@BoltzmannBrain 

I fixes the problem of the temporal pooler region. You should be able to initialize it now. 

There were two problems. The first problem is because I forgot to pass in the poolerType when calling _getParentSpatialPoolerClass. Sorry about the bug.

To get it running, I also reverted an earlier commit that removed sequenceIDIn from the BaseSpec of TemporalPoolerRegion. 
https://github.com/numenta/nupic.research/commit/c13e74d34a1ce048711217a29c9fb920402b2f25#diff-29ead8af5de93cf140f7faf013f3002bL362

Why do we eliminate the sequenceIDIn information from TemporalPooler? If we really need to do that, we also need to change this
https://github.com/numenta/nupic.research/blob/master/htmresearch/frameworks/classification/classification_network.py#L326

### DIFF
--- a/htmresearch/regions/TemporalPoolerRegion.py
+++ b/htmresearch/regions/TemporalPoolerRegion.py
@@ -35,9 +35,9 @@ class MonitoredUnionTemporalPooler(UnionTemporalPoolerMonitorMixin,
 
 
 def _getPoolerClass(name):
-    if name=="union":
+    if name == "union":
       return UnionTemporalPooler
-    elif name=="unionMonitored":
+    elif name == "unionMonitored":
       return MonitoredUnionTemporalPooler
     else:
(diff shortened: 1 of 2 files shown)


BoltzmannBrain:
Thanks @ywcui1990 
There was some confusion as to where to implement the sequenceId links, here and in nupic. All regions should have a sequenceId input, and the network factory should create this link from the sensor to all regions. @SaganBolliger has PRs open to this effect:
- https://github.com/numenta/nupic/pull/2770
- https://github.com/numenta/nupic.research/pull/370

BoltzmannBrain:
:+1:

PR #388, opened 2015-12-22
Misc. improvements to buckets experiment

BoltzmannBrain:


BoltzmannBrain:
@subutai this has a few fixups for the buckets experiment, mainly adding the ability to specify the distance metric used in the classifier (which is useful for multiple experiments and Imbu).

subutai:
:+1:

PR #394, opened 2015-12-31
IMBU-53 Constructor argument for user-defined cache root, defaults to…

oxtopus:
… original calculated location if not specified


BoltzmannBrain:
Good change, but please implement this in the other models as well: ClassificationModelWindows, ClassificationModelHTM, and ClassificationModelEndpoint.

PR #410, opened 2016-01-11
[RES-140] Simple union pooler Implementation

ywcui1990:
@subutai @BoltzmannBrain 

* Implement a simple union pooler that computes a union of the last N SDRs
* Add test for simple union pooler

### DIFF
--- a/htmresearch/algorithms/simple_union_pooler.py
+++ b/htmresearch/algorithms/simple_union_pooler.py
@@ -46,56 +46,81 @@ class SimpleUnionPooler(object):
 		@param historyLength: The union window length. For a union of the last
 		10 steps, use historyLength=10
 		"""
-		inputDimensions = numpy.array(inputDimensions, ndmin=1)
 
-		self._numInputs = inputDimensions.prod()
+		self._inputDimensions = inputDimensions
 		self._historyLength = historyLength
+		self._numInputs = 1
+		for d in inputDimensions:
+			self._numInputs *= d
 
 		# Current union SDR; the output of the union pooler algorithm
-		self._unionSDR = numpy.zeros(shape=inputDimensions, dtype=UINT_DTYPE)
+		self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
 
 		# Indices of currently active cells
-		self._activeCells = set()
-
-		# Activation time of all cells
-		self._activationTimer = numpy.zeros(self._numInputs, dtype=REAL_DTYPE)
+		self._activeCellsHistory = []
 
 
 	def reset(self):
-		self._unionSDR = numpy.array([], dtype=UINT_DTYPE)
-		self._activeCells = set()
-		self._activationTimer = numpy.zeros(self._numInputs, dtype=REAL_DTYPE)
-
-
-	def compute(self, inputVector):
 		"""
-		Computes one cycle of the Union Pooler algorithm.
-		@param inputVector: A numpy array of 0's and 1's that comprises the
-		input to the union pooler. The array will be treated as a one dimensional array
+		Reset Union Pooler, clear active cell history
 		"""
-
-		if not isinstance(inputVector, numpy.ndarray):
-			raise TypeError("Input vector must be a numpy array, not %s" %
-											str(type(inputVector)))
-
-		if inputVector.size != self._numInputs:
-			raise ValueError(
-					"Input vector dimensions don't match. Expecting %s but got %s" % (
-							inputVector.size, self._numInputs))
-
-		inputVector = numpy.array(inputVector, dtype=REAL_DTYPE)
-		inputVector.reshape(-1)
-
-		activeBits = set(numpy.where(inputVector)[0])
-
-		self._activeCells = self._activeCells | activeBits
-		self._activationTimer[list(self._activeCells)] += 1
-
-		cellsPastHistoryLength = set(numpy.where(self._activationTimer > self._historyLength)[0])
-		self._activeCells = self._activeCells - cellsPastHistoryLength
+		self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
+		self._activeCellsHistory = []
 
 
-	def getUnionSDR(self):
-		self._unionSDR = numpy.zeros(shape=(self._numInputs,), dtype=UINT_DTYPE)
-		self._unionSDR[list(self._activeCells)] = 1
+	def compute(self, activeCells):
+		"""
+		Computes one cycle of the Union Pooler algorithm. Return the union SDR
+		Parameters:
+		----------------------------
+		@param activeCells: A list that stores indices of active cells
+		"""
+		self._activeCellsHistory.append(activeCells)
+		if len(self._activeCellsHistory) > self._historyLength:
+			self._activeCellsHistory.pop(0)
+
+		self.createUnionSDR()
 		return self._unionSDR
+
+
+	def createUnionSDR(self):
+		"""
+		Create union SDR from a history of active cells
+		"""
+		self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
+		for i in self._activeCellsHistory:
+			self._unionSDR[i] = 1
+
+
+	def unionIntoArray(self, inputVector, outputVector):
+		"""
+		Create a union of the inputVector and copy the result into the output Vector
+		Parameters:
+		----------------------------
+		@param inputVector: The inputVector can be either a full numpy array containing
+		0's and 1's, or a list of non-zero entry indices
+		@param outputVector: A numpy array that matches the inputDimensions of the
+		union pooler.
+		"""
+		if isinstance(inputVector, numpy.ndarray):
+			if inputVector.size == self._numInputs:
+				activeBits = numpy.where(inputVector)[0]
+			else:
+				raise ValueError(
+					"Input vector dimensions don't match. Expecting %s but got %s" % (
+						self._numInputs, inputVector.size))
+		elif isinstance(inputVector, list):
+			if max(inputVector) >= self._numInputs:
+				raise ValueError(
+					"Non-zero entry indice exceeds input dimension of union pooler. "
+					"Expecting %s but got %s" % (self._numInputs, max(inputVector)))
+			else:
+				activeBits = inputVector
+		else:
+			raise ValueError("Unsuported input types")
+
+		self.compute(activeBits)
+
+		outputVector[:] = 0
+		for i in self._activeCellsHistory:
+			outputVector[i] = 1
--- a/htmresearch/algorithms/simple_union_pooler.py
+++ b/htmresearch/algorithms/simple_union_pooler.py
@@ -20,11 +20,6 @@
 # ----------------------------------------------------------------------
 
 import numpy
-from nupic.bindings.math import GetNTAReal
-
-REAL_DTYPE = GetNTAReal()
-UINT_DTYPE = "uint32"
-
 
 class SimpleUnionPooler(object):
   """
@@ -33,27 +28,18 @@ class SimpleUnionPooler(object):
   """
 
   def __init__(self,
-               inputDimensions=(2048,),
+               numInputs=2048,
                historyLength=10):
     """
     Parameters:
     ----------------------------
-    @param inputDimensions:
-    A sequence representing the dimensions of the input vector. Format is
-    (height, width, depth, ...), where each value represents the size of the
-    dimension.  For a topology of one dimension with 100 inputs use 100, or
-    (100,). For a two dimensional topology of 10x5 use (10,5).
-
+    @param numInputs: The length of the input SDRs
     @param historyLength: The union window length. For a union of the last
     10 steps, use historyLength=10
     """
 
-    self._inputDimensions = inputDimensions
     self._historyLength = historyLength
-    self._numInputs = 1
-    for d in inputDimensions:
-      self._numInputs *= d
-
+    self._numInputs = numInputs
     self.reset()
 
 
@@ -61,11 +47,11 @@ class SimpleUnionPooler(object):
     """
     Reset Union Pooler, clear active cell history
     """
-    self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
+    self._unionSDR = numpy.zeros(shape=(self._numInputs,))
     self._activeCellsHistory = []
 
 
-  def compute(self, activeCells):
+  def updateHistory(self, activeCells):
     """
     Computes one cycle of the Union Pooler algorithm. Return the union SDR
     Parameters:
@@ -76,17 +62,11 @@ class SimpleUnionPooler(object):
     if len(self._activeCellsHistory) > self._historyLength:
       self._activeCellsHistory.pop(0)
 
-    self.createUnionSDR()
-    return self._unionSDR
-
-
-  def createUnionSDR(self):
-    """
-    Create union SDR from a history of active cells
-    """
-    self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
+    self._unionSDR = numpy.zeros(shape=(self._numInputs,))
     for i in self._activeCellsHistory:
       self._unionSDR[i] = 1
+
+    return self._unionSDR
 
 
   def unionIntoArray(self, inputVector, outputVector):
@@ -116,11 +96,14 @@ class SimpleUnionPooler(object):
     else:
       raise TypeError("Unsuported input types")
 
-    self.compute(activeBits)
+    if len(outputVector) != self._numInputs:
+      raise ValueError(
+        "Output vector dimension does match dimension of union pooler "
+        "Expecting %s but got %s" % (self._numInputs, len(outputVector)))
 
-    outputVector[:] = 0
-    for i in self._activeCellsHistory:
-      outputVector[i] = 1
+    unionSDR = self.updateHistory(activeBits)
+    for i in xrange(len(unionSDR)):
+      outputVector[i] = unionSDR[i]
 
 
   def getSparsity(self):
--- a/htmresearch/algorithms/simple_union_pooler.py
+++ b/htmresearch/algorithms/simple_union_pooler.py
@@ -76,7 +76,7 @@ class SimpleUnionPooler(object):
     ----------------------------
     @param inputVector: The inputVector can be either a full numpy array
     containing 0's and 1's, or a list of non-zero entry indices
-    @param outputVector: A numpy array that matches the inputDimensions of the
+    @param outputVector: A numpy array that matches the length of the
     union pooler.
     """
     if isinstance(inputVector, numpy.ndarray):
@@ -102,6 +102,7 @@ class SimpleUnionPooler(object):
         "Expecting %s but got %s" % (self._numInputs, len(outputVector)))
 
     unionSDR = self.updateHistory(activeBits)
+
     for i in xrange(len(unionSDR)):
       outputVector[i] = unionSDR[i]
 
--- a/htmresearch/algorithms/simple_union_pooler.py
+++ b/htmresearch/algorithms/simple_union_pooler.py
@@ -1,6 +1,6 @@
 # ----------------------------------------------------------------------
 # Numenta Platform for Intelligent Computing (NuPIC)
-# Copyright (C) 2015, Numenta, Inc.  Unless you have an agreement
+# Copyright (C) 2016, Numenta, Inc.  Unless you have an agreement
 # with Numenta, Inc., for a separate license for this software code, the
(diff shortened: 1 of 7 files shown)


htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+		self._unionSDR = numpy.zeros(shape=inputDimensions, dtype=UINT_DTYPE)
+
+		# Indices of currently active cells
+		self._activeCells = set()
+
+		# Activation time of all cells
+		self._activationTimer = numpy.zeros(self._numInputs, dtype=REAL_DTYPE)
+
+
+	def reset(self):
```
subutai: docstring?

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+		@param inputDimensions:
+			A sequence representing the dimensions of the input vector. Format is
+			(height, width, depth, ...), where each value represents the size of the
+			dimension.  For a topology of one dimension with 100 inputs use 100, or
+			(100,). For a two dimensional topology of 10x5 use (10,5).
+
+		@param historyLength: The union window length. For a union of the last
+		10 steps, use historyLength=10
+		"""
+		inputDimensions = numpy.array(inputDimensions, ndmin=1)
```
subutai: This is very strange - why do you need to create an array that does nothing useful?

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+
+class SimpleUnionPooler(object):
+	"""
+	Experimental Simple Union Pooler Python Implementation.
+	The simple union pooler computes a union of the last N SDRs
+	"""
+
+
+	def __init__(self,
+							 inputDimensions=(2048,),
```
subutai: Do we need to support multiple dimensions? Why not just a single array?

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+			(height, width, depth, ...), where each value represents the size of the
+			dimension.  For a topology of one dimension with 100 inputs use 100, or
+			(100,). For a two dimensional topology of 10x5 use (10,5).
+
+		@param historyLength: The union window length. For a union of the last
+		10 steps, use historyLength=10
+		"""
+		inputDimensions = numpy.array(inputDimensions, ndmin=1)
+
+		self._numInputs = inputDimensions.prod()
```
subutai: Could just be:
```
self._numInputs = 1
for d in inputDimensions: self._numInputs *= d
```

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+		self._activationTimer = numpy.zeros(self._numInputs, dtype=REAL_DTYPE)
+
+
+	def reset(self):
+		self._unionSDR = numpy.array([], dtype=UINT_DTYPE)
+		self._activeCells = set()
+		self._activationTimer = numpy.zeros(self._numInputs, dtype=REAL_DTYPE)
+
+
+	def compute(self, inputVector):
```
subutai: The design of this function is very confusing to me. How is a client supposed to use this class? The client code should not access internal protected variables and the tests never actually call getUnionSDR. I would expect the compute function to return the union SDR.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+		activeBits = set(numpy.where(inputVector)[0])
+
+		self._activeCells = self._activeCells | activeBits
+		self._activationTimer[list(self._activeCells)] += 1
+
+		cellsPastHistoryLength = set(numpy.where(self._activationTimer > self._historyLength)[0])
+		self._activeCells = self._activeCells - cellsPastHistoryLength
+
+
+	def getUnionSDR(self):
```
subutai: Documentation? Also this method is not used anywhere.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,101 @@
...
+
+		if inputVector.size != self._numInputs:
+			raise ValueError(
+					"Input vector dimensions don't match. Expecting %s but got %s" % (
+							inputVector.size, self._numInputs))
+
+		inputVector = numpy.array(inputVector, dtype=REAL_DTYPE)
+		inputVector.reshape(-1)
+
+		activeBits = set(numpy.where(inputVector)[0])
```
subutai: set operations are very slow in Python. Why did you choose to use a set here?

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,75 @@
...
+			npt.assert_allclose(self.unionPooler._activationTimer[numpy.array(activeCells[i])],
+			                    len(activeCells)-i)
+
+	def testHistoryLength(self):
+		self.unionPooler = SimpleUnionPooler(inputDimensions=(2048,),
+		                                     historyLength=2)
+		activeCells = []
+		activeCells.append([1, 3, 4])
+		activeCells.append([101, 302, 405])
+		activeCells.append([240, 903, 858])
```
subutai: The tests should include some tricky situations. For example, what if the first and second one both had bit 42 on? In this case it should not be removed from the union.

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+		for i in xrange(len(activeCells)):
+			inputVector = numpy.zeros(shape=(2048,))
+			inputVector[numpy.array(activeCells[i])] = 1
+			self.unionPooler.compute(inputVector)
+
+		activeCellsUnion = set()
+		for i in xrange(1, len(activeCells)):
```
subutai: I think this would be simpler to understand and better from a testing standpoint if you just set activeCellsUnion explicitly to (101, 302, 405, 240, 903, 858) rather than putting in yet another loop that could also contain bugs. It's also fewer lines of code!

BoltzmannBrain:
@ywcui1990 it looks like Subutai gave you a thorough review, so I'll take a look after you respond to his comments. But also please set your indents to 2 spaces and line wraps at 80 characters :smile:

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,75 @@
...
+			self.unionPooler.compute(inputVector)
+
+		activeCellsUnion = set()
+		for i in xrange(len(activeCells)):
+			activeCellsUnion = activeCellsUnion | set(activeCells[i])
+		self.assertSetEqual(self.unionPooler._activeCells, activeCellsUnion)
+
+		for i in xrange(len(activeCells)):
+			npt.assert_allclose(self.unionPooler._activationTimer[numpy.array(activeCells[i])],
+			                    len(activeCells)-i)
```
BoltzmannBrain: Why not use the standard unittest [assertAlmostEqual()](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertAlmostEqual)?

ywcui1990:
I have incorporated Subutai's comments and suggestions. Please review.
@subutai @BoltzmannBrain

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+		outputVector = numpy.zeros(shape=(2048,))
+		for i in xrange(len(activeCells)):
+			self.unionPooler.unionIntoArray(activeCells[i], outputVector)
+
+		activeCellsUnion = []
+		for i in xrange(len(activeCells)):
+			activeCellsUnion += activeCells[i]
+
+		self.assertSetEqual(set(numpy.where(outputVector)[0]),
```
BoltzmannBrain: Why does `outputVector` need to be a set? Shouldn't that be handled w/in the SimpleUnionPooler?

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,73 @@
...
+		activeCells.append([1, 3, 4])
+		activeCells.append([101, 302, 405])
+
+		outputVector = numpy.zeros(shape=(2048,))
+		for i in xrange(len(activeCells)):
+			self.unionPooler.unionIntoArray(activeCells[i], outputVector)
+
+		activeCellsUnion = []
+		for i in xrange(len(activeCells)):
+			activeCellsUnion += activeCells[i]
```
BoltzmannBrain:
```
for cells in activeCells:
  activeCellsUnion += cells
```

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+import unittest
+import numpy
+from htmresearch.algorithms.simple_union_pooler import SimpleUnionPooler
+
+REAL_DTYPE = numpy.float32
+
+
+class SimpleUnionPoolerTest(unittest.TestCase):
+	def setUp(self):
```
BoltzmannBrain: indents are still off in this file

htmresearch/frameworks/union_temporal_pooling/activation/excite_functions/excite_functions_all.py
```diff
@@ -62,7 +62,7 @@ def excite(self, currentActivation, inputs):
     @param inputs            (numpy array) inputs for each cell
     """
 
-    currentActivation = self._minValue + (self._maxValue - self._minValue) / \
+    currentActivation += self._minValue + (self._maxValue - self._minValue) / \
```
BoltzmannBrain: use implicit continuation with parenthesis rather than backslashes. If the equation gets too complex to follow, split it up into several calculations.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,127 @@
...
+            self._numInputs, inputVector.size))
+    elif isinstance(inputVector, list):
+      if max(inputVector) >= self._numInputs:
+        raise ValueError(
+          "Non-zero entry indice exceeds input dimension of union pooler. "
+          "Expecting %s but got %s" % (self._numInputs, max(inputVector)))
+      else:
+        activeBits = inputVector
+    else:
+      raise ValueError("Unsuported input types")
```
BoltzmannBrain: Unsupported
Use a TypeError

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,127 @@
...
+      if inputVector.size == self._numInputs:
+        activeBits = numpy.where(inputVector)[0]
+      else:
+        raise ValueError(
+          "Input vector dimensions don't match. Expecting %s but got %s" % (
+            self._numInputs, inputVector.size))
+    elif isinstance(inputVector, list):
+      if max(inputVector) >= self._numInputs:
+        raise ValueError(
+          "Non-zero entry indice exceeds input dimension of union pooler. "
```
BoltzmannBrain: indices

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,127 @@
...
+    self._historyLength = historyLength
+    self._numInputs = 1
+    for d in inputDimensions:
+      self._numInputs *= d
+
+    # Current union SDR; the output of the union pooler algorithm
+    self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
+
+    # Indices of currently active cells
+    self._activeCellsHistory = []
```
BoltzmannBrain: Cleaner would be to replace these few lines with `self.reset()`.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,127 @@
...
+import numpy
+from nupic.bindings.math import GetNTAReal
+
+REAL_DTYPE = GetNTAReal()
+UINT_DTYPE = "uint32"
+
+
+class SimpleUnionPooler(object):
+  """
+	Experimental Simple Union Pooler Python Implementation.
```
BoltzmannBrain: your docstring tabs are off here and throughout

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,127 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2015, Numenta, Inc.  Unless you have an agreement
```
BoltzmannBrain: 2016 (here and throughout)

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,73 @@
...
+		for i in xrange(len(activeCells)):
+			inputVector = numpy.zeros(shape=(2048,))
+			inputVector[numpy.array(activeCells[i])] = 1
+			self.unionPooler.unionIntoArray(activeCells[i], outputVector)
+
+		self.assertSetEqual(set(numpy.where(outputVector)[0]),
+		                    set(activeCellsUnion))
+
+
+if __name__ == '__main__':
```
BoltzmannBrain: double quotes please

BoltzmannBrain:
@ywcui1990 done with review. I think an important test case to include is repeating active cells -- i.e. `[13, 42, 58, 198]` is added to `activeCells`, followed later by `[55, 72, 198, 272]`.

Also, a `getSparsity()` method may be useful.

tests/union_pooling/simple_union_pooler_test.py
ywcui1990: outputVector is a numpy array. I computed a set of non-zero indices of numpy array to verify that outputVector has been computed correctly.

ywcui1990:
@BoltzmannBrain Thanks for the review. I added the test and the getSparsity method.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    self._activeCellsHistory.append(activeCells)
+    if len(self._activeCellsHistory) > self._historyLength:
+      self._activeCellsHistory.pop(0)
+
+    self.createUnionSDR()
+    return self._unionSDR
+
+
+  def createUnionSDR(self):
```
subutai: This method is not used anywhere else and is just three lines. Why not just move these three lines into compute()? Is there an advantage to having this function?

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+      if max(inputVector) >= self._numInputs:
+        raise ValueError(
+          "Non-zero entry indices exceed input dimension of union pooler. "
+          "Expecting %s but got %s" % (self._numInputs, max(inputVector)))
+      else:
+        activeBits = inputVector
+    else:
+      raise TypeError("Unsuported input types")
+
+    self.compute(activeBits)
```
subutai: The return value of compute is not used and thus the unioning operation is done twice!  I'm confused.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+
+  def reset(self):
+    """
+    Reset Union Pooler, clear active cell history
+    """
+    self._unionSDR = numpy.zeros(shape=self._inputDimensions, dtype=UINT_DTYPE)
+    self._activeCellsHistory = []
+
+
+  def compute(self, activeCells):
```
subutai: Just call this function updateHistory(self, activeCells) and don't create the union - it's not used anywhere!!

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+import numpy
+from nupic.bindings.math import GetNTAReal
+
+REAL_DTYPE = GetNTAReal()
```
subutai: Don't need REAL_DTYPE and UINT_DTYPE. The loop in unionIntoArray() will work with any data type.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class SimpleUnionPooler(object):
+  """
+  Experimental Simple Union Pooler Python Implementation.
+  The simple union pooler computes a union of the last N SDRs
+  """
+
+  def __init__(self,
+               inputDimensions=(2048,),
```
subutai: Do we really need to support multi-dimensional inputs here? Why? The inputs to regions are always flat arrays. Even Cortical IO uses indices computed from a flat array.

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,85 @@
...
+                                         historyLength=10)
+
+
+  def testUnionCompute(self):
+    activeCells = []
+    activeCells.append([1, 3, 4])
+    activeCells.append([101, 302, 405])
+    activeCellsUnion = [1, 3, 4, 101, 302, 405]
+
+    outputVector = numpy.zeros(shape=(2048,))
```
subutai: If you are going to support multi-dimensional arrays, you should test with multi-dimensional arrays.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+    Parameters:
+    ----------------------------
+    @param inputVector: The inputVector can be either a full numpy array
+    containing 0's and 1's, or a list of non-zero entry indices
+    @param outputVector: A numpy array that matches the inputDimensions of the
+    union pooler.
+    """
+    if isinstance(inputVector, numpy.ndarray):
+      if inputVector.size == self._numInputs:
+        activeBits = numpy.where(inputVector)[0]
```
subutai: I'm not convinced these statements and loops work correctly for multi-dimensional arrays. Need tests. For example, you also need a test whether the shapes match - not sufficient to match size.

tests/union_pooling/simple_union_pooler_test.py
```diff
@@ -0,0 +1,85 @@
...
+    activeCells.append([55, 72, 198, 272])
+    activeCellsUnion = [13, 42, 58, 198, 55, 72, 272]
+
+    outputVector = numpy.zeros(shape=(2048,))
+    for i in xrange(len(activeCells)):
+      self.unionPooler.unionIntoArray(activeCells[i], outputVector)
+
+    self.assertSetEqual(set(numpy.where(outputVector)[0]),
+                        set(activeCellsUnion))
+
```
subutai: Need to include test cases that test negative cases. For example the situation where you don't send an array with the correct dimensions. The test should ensure that an exception is raised (use the check assertRaises()). See for example https://github.com/numenta/nupic/blob/master/tests/unit/nupic/algorithms/knn_classifier_test.py#L249

subutai:
@ywcui1990 I left some comments - the implementation makes no sense to me!

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,131 @@
...
+    for i in self._activeCellsHistory:
+      self._unionSDR[i] = 1
+
+
+  def unionIntoArray(self, inputVector, outputVector):
+    """
+    Create a union of the inputVector and copy the result into the outputVector
+    Parameters:
+    ----------------------------
+    @param inputVector: The inputVector can be either a full numpy array
```
subutai: I like this idea - makes the method very flexible.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,115 @@
...
+      raise TypeError("Unsuported input types")
+
+    if len(outputVector) != self._numInputs:
+      raise ValueError(
+        "Output vector dimension does match dimension of union pooler "
+        "Expecting %s but got %s" % (self._numInputs, len(outputVector)))
+
+    unionSDR = self.updateHistory(activeBits)
+
+    for i in xrange(len(unionSDR)):
```
subutai: You do need to do outputVector[:] = 0 first.  I think you used to have that before.

subutai: Never mind - I see that you are doing a full vector copy.

htmresearch/algorithms/simple_union_pooler.py
```diff
@@ -0,0 +1,115 @@
...
+      outputVector[i] = unionSDR[i]
```
subutai: I believe numpy.copyto(outputVector,unionSDR,casting="unsafe") would be faster than doing this loop.

subutai:
I think this version is much better. Left one speedup comment. I think a test for the reset function would be useful to have. Otherwise :+1:

PR #414, opened 2016-01-12
Map original text to processed tokens

BoltzmannBrain:


BoltzmannBrain:
@subutai please review.
These changes get the word-to-token mappings we discussed (for Imbu).

htmresearch/frameworks/nlp/classification_model.py
```diff
@@ -188,25 +188,28 @@ def inferDocument(self, document, returnDetailedResults=False,
...
     if self.filterText:
-      sample = TextPreprocess().tokenize(preprocessedText,
+      sample, mapping = TextPreprocess().tokenizeAndFilter(inputText,
                                          ignoreCommon=100,
                                          removeStrings=["[identifier deleted]"],
                                          correctSpell=True)
     else:
-      sample = TextPreprocess().tokenize(preprocessedText)
+      sample, mapping = TextPreprocess().tokenizeAndFilter(inputText)
+      # print
```
subutai: Remove commented out print statements

htmresearch/support/text_preprocess.py
```diff
@@ -68,12 +68,20 @@ def __init__(self,
...
-    corpusPath = os.path.abspath(os.path.join(
-      os.path.dirname(__file__), "../..", "projects/nlp/data/etc", corpusSource))
     try:
+      if os.path.isfile(corpusSource):
+        # absolute path
+        corpusPath = corpusSource
+      else:
+        # relative path
+        corpusPath = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                  "../../projects/nlp/data/etc",
```
subutai: Not sure if we really care but this code is not portable to Windows. In order to be portable it should be something like:

```
os.path.join(os.path.dirname(__file__),
                                                  "..","..","projects/nlp/data/etc", corpusSource)
```

htmresearch/support/csv_helper.py
```diff
@@ -51,17 +51,23 @@ def readCSV(csvFile, numLabels=0):
...
         idIdx = 0
-      labelIdx = range(sampleIdx + 1, sampleIdx + 1 + numLabels)
-
+      
       dataDict = OrderedDict()
-      for lineNumber, line in enumerate(reader):
-        dataDict[lineNumber] = (line[sampleIdx],
-                                [line[i] for i in labelIdx if line[i]],
-                                line[idIdx])
+
```
subutai: Why do we need an OrderedDict here? Wouldn't a normal dict suffice?

htmresearch/support/text_preprocess.py
```diff
@@ -147,50 +156,97 @@ def readExpansionFile(filename, suffixes=None):
...
+    @return mapping        (dict)   Keys are the indices of the input words.
+        Values are two-tuples, where the first item is the original string word
+        (w/ punctuation) and the second item is a list of string tokens. This
+        format is necessary to account for when there are more tokens than words
+        and vice-versa, and can't use words as keys b/c repeats would get
+        overwritten.
     """
+    originalWords = text.split(" ")
+    processedWords = []
+    mapping  = OrderedDict()
```
subutai: Why OrderedDict? Wouldn't a regular dict suffice?

subutai:
Generally looks good. I'm not convinced this particular mapping is the most convenient return type. How would a client use this while training a model? We would be passing in one token a time to the model, along with the index associated with that token. That is not readily available in the mapping.

htmresearch/frameworks/nlp/classification_model.py
BoltzmannBrain: ah sorry

htmresearch/support/csv_helper.py
BoltzmannBrain: It was necessary in the previous versions of the NLP models. So yes, a normal dict would be better (simpler).

htmresearch/support/text_preprocess.py
BoltzmannBrain: I thought it could be useful to have the order preserved, but it would be simpler to use a normal dict. If order is something we need later we can always change it then.

subutai: Order would be preserved in a normal dict too, since the key is the index!

BoltzmannBrain:
@subutai please take a look now. This implements a more useful mapping with the indexing scheme we discussed.

In addition to tests/nlp/integration/text_mapping_test.py, you can check out the mappings by running Imbu: `python htmresearch/frameworks/nlp/imbu.py --dataPath project/nlp/data/sample_reviews/sample_reviews.csv -m HTMNetwork`

Fixing the Keywords detailed inference method, I realized there would be a lot of duplicate code so I just added an if-statement for exact matching in the base implementation.

I tried squashing the merge master commits through rebasing but it wouldn't work :unamused:

htmresearch/frameworks/nlp/classification_model.py
```diff
@@ -88,17 +89,17 @@ def trainDocument(self, document, labels, sampleId):
...
+    @param document   (str)  The text token to train on.
     @param labels     (list) A list of one or more integer labels associated
                              with this token.
     @param sampleId   (int)  An integer ID associated with the entire document.
-
     """
-    # Default implementation, may be overridden
+    # Default implementation, may be overridden. Here we ignore the token-word
+    # mappings produced by the tokenizer.
     assert (sampleId is not None), "Must pass in a sampleId"
```
subutai: Should be replaced with wordID here and elsewhere.

htmresearch/frameworks/nlp/classify_htm.py
```diff
@@ -104,6 +89,32 @@ def trainToken(self, token, labels, sampleId, reset=0):
...
+    been trained.
+
+    @param document   (str)  The text token to train on.
+    @param labels     (list) A list of one or more integer labels associated
+                             with this token.
+    @param sampleId   (int)  An integer ID associated with the entire document.
+    """
+    # Train on each token with an id that points to its corresponding word in
+    # the original text. This id is generated by multiplying the document's id
+    # by an indexing factor (e.g. 1000), and then adding the mapping index (from
```
subutai: This won't let you associate a single id with an entire document. 

I think the indexing logic should be in the client, not the model. It's very application specific. For example if we just want to classify documents, we wouldn't do this. The model should just store the IDs it is given.

htmresearch/frameworks/nlp/classify_keywords.py
```diff
@@ -72,7 +73,33 @@ def trainToken(self, token, labels, sampleId, reset=0):
...
+    This routine will tokenize the document and train the model on these tokens
+    using the given labels and ID. A reset will be issued after the document has
+    been trained.
+
+    @param document   (str)  The text token to train on.
+    @param labels     (list) A list of one or more integer labels associated
+                             with this token.
+    @param sampleId   (int)  An integer ID associated with the entire document.
+    """
+    # Train on each token with an id that points to its corresponding word in
```
subutai: See above comment on where this logic should be.

htmresearch/frameworks/nlp/classify_htm.py
```diff
@@ -104,6 +89,32 @@ def trainToken(self, token, labels, sampleId, reset=0):
...
+    # the tokenizer); e.g., the tenth token of the third document will have
```
subutai: In other words, if you want to do windows, the client would have to use the trainToken API directly. The model code doesn't really need to change.

subutai:
@BoltzmannBrain Just had one major concern, which is that we shouldn't pollute the model hierarchy with the indexing logic. That is very application specific and a client responsibility - the client should instead just use trainToken.

htmresearch/frameworks/nlp/classify_htm.py
BoltzmannBrain: Yes that's better, but then the client would become a little more complicated by calling `trainToken()` for some models and `trainDocument()` for others.

htmresearch/frameworks/nlp/classify_htm.py
BoltzmannBrain: Ah, should've read this comment before replying above :smile:

htmresearch/frameworks/nlp/classify_htm.py
subutai: It can still call trainToken for all models - they all support that API.

BoltzmannBrain:
@subutai please take a look now.

subutai:
:+1:

PR #425, opened 2016-01-19
[RES-140] Add simple union pooler to TemporalPoolerRegion

ywcui1990:
@subutai @BoltzmannBrain 

* Add the simple union pooler to TemporalPoolerRegion
* Add a network config file that uses TemporalPoolerRegion with simple union pooler


subutai:
@ywcui1990 There are tons of problems in the basic TemporalPoolerRegion implementation, so I think this is fine for now. Thanks for checking in the JSON file. Can you also create a version of sensor_knn.json with simple union pooling, and then verify it works with hello_classification_model.py?  You should get decent answers (90% accuracy). 

I don't believe the save/load command will work due to existing TemporalPoolerRegion bugs, so you may need to comment that out when you test.

ywcui1990:
@subutai 

I have verified that the adding the TemporalPoolerRegion still gives 90% accuracy on the hello_classification_model example. The save/load command also seem to work, as I got the same accuracy when testing after de-serialization

PR #428, opened 2016-01-20
Add Echo state network result to the continuous sequence task

ywcui1990:
@subutai 
Summary

* Run the NYC taxi data with Echo State Network (requires an external Matlab toolbox)
* Combine ESN results with the rest of the algorithms
* ESN performance is between ARIMA and LSTM/TM on both NRMSE and MAPE metric

model_performance_summary_alternative


subutai:
Thanks! Looks nice. I didn't really look through the code. :+1:

floybix:
Keeping in mind that KNN outperforms all of them on this problem...

ywcui1990:
Thanks @floybix I will look into KNN next.

PR #432, opened 2016-01-21
Sort NLP inference votes

BoltzmannBrain:
@subutai please review.


BoltzmannBrain:
@subutai ping
I'd like to merge this now to avoid conflicts with #414.

htmresearch/frameworks/nlp/classification_model.py
```diff
@@ -416,3 +420,19 @@ def _inferDocumentDetailed(self, tokenList, sortResults=True):
 
     else:
       return categoryVotes, sampleIdList, distancetoSampleIds
+
+
+  @staticmethod
+  def _sortArray(array, seed=42):
```
subutai: Since the seed is always fixed here, the tie breaking will not be random - it will be a fixed order each time. Is that what you want?

htmresearch/frameworks/nlp/classification_model.py
```diff
@@ -416,3 +420,19 @@ def _inferDocumentDetailed(self, tokenList, sortResults=True):
...
+
+    @param array (numpy array)    Array of values to be sorted.
+    @param seed (int)             Seed the random number generator.
+
+    @return   (numpy array)       Sorted array indices, where the sort order is
+                                  greatest to least.
+    """
+    numpy.random.seed(seed)
+    randomValues = numpy.random.random(array.size)
+    return numpy.lexsort((randomValues, array))[::-1]
```
subutai: I'm not familiar with lexsort - does it only use the randomValues for tie breaking?

htmresearch/frameworks/nlp/classification_model.py
```diff
@@ -181,7 +181,9 @@ def inferDocument(self, document, returnDetailedResults=False,
                                      sortResults=False)
 
       if votes.sum() > 0:
-        categoryVotes[votes.argmax()] += 1
+        # Increment the most likely category, breaking ties in a random fashion
```
subutai: It should only sort if sortResults=True

subutai:
Left a few comments. Definitely don't need to wait for me to merge.

htmresearch/frameworks/nlp/classification_model.py
BoltzmannBrain: Yes; `numpy.lexsort((b,a))` means sort by a, then b.

htmresearch/frameworks/nlp/classification_model.py
BoltzmannBrain: Yes, we want ties between items to be broken without always giving precedence to earlier items, but still get repeatable results.

htmresearch/frameworks/nlp/classification_model.py
BoltzmannBrain: `sortResults` is for `_inferDocumentDetailed()`. The sorting here is to the same effect as `votes.argmax()` -- to get the maximum category so we can increment its votes. I'll change the naming so it's less deceiving.

BoltzmannBrain:
Thanks @subutai :)

PR #438, opened 2016-01-25
Imbu use multiple HTM Network configs

BoltzmannBrain:
@oxtopus please review.

### DIFF
--- a/htmresearch/frameworks/nlp/imbu.py
+++ b/htmresearch/frameworks/nlp/imbu.py
@@ -68,7 +68,7 @@ def _loadNetworkConfig(jsonName=None):
   """ Load network config by calculating path relative to this file, and load
   with htmresearch.frameworks.nlp.model_factory.getNetworkConfig()
   """
-  if jsonName is None:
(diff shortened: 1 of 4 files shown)


htmresearch/frameworks/nlp/imbu.py
```diff
@@ -64,10 +64,13 @@ def __init__(self, exc):
 
 
 
-def _loadNetworkConfig():
+def _loadNetworkConfig(jsonName=None):
   """ Load network config by calculating path relative to this file, and load
   with htmresearch.frameworks.nlp.model_factory.getNetworkConfig()
   """
+  if jsonName is None:
```
oxtopus: `if not jsonName` would catch `None` as well as empty string or anything the resolves to `False`

oxtopus:
:+1:

htmresearch/frameworks/nlp/imbu.py
BoltzmannBrain: Yes that's better, thanks!

PR #440, opened 2016-01-26
Misc. changes in order to run more NLP models in Imbu

BoltzmannBrain:


htmresearch/frameworks/nlp/model_factory.py
```diff
@@ -47,15 +47,23 @@ class ClassificationModelTypes(object):
...
+  ciodoc = ClassificationModelFingerprint
+  cioword = ClassificationModelFingerprint
+
   Keywords = ClassificationModelKeywords
+  keywords = ClassificationModelKeywords
+
   HTMNetwork = ClassificationModelHTM
+  htm = ClassificationModelHTM
+  htm1 = ClassificationModelHTM
+  htm2 = ClassificationModelHTM
```
subutai: I'm confused as to the purpose of these - ???

BoltzmannBrain: WIP :)

BoltzmannBrain:
@subutai these changes are needed to run Imbu. 

1. We want to run multiple HTM Network saved models at the same time, which would need different directory names -- e.g., "htm_sensor_simple_tp_knn/" and "htm_sensor_knn/". The Imbu model runner uses these directory names as model names, which are mapped to the corresponding class via `ClassificationModelTypes`. There are alternative solutions to listing additional identifiers in `ClassificationModelTypes`, but this is the most straightforward, serves as a good means of self-documenting, and forces us to stay concise with the number of identifiers we allow.
2. The TP region requires matplotlib.

PR #443, opened 2016-01-27
Changes FaultyTemporalMemory and Implements FaultyTemporalMemoryShim

ywcui1990:
@subutai 
As we discussed today. I have made some changes while trying to get the FaultyTM working with the Discrete Sequence Prediction task. 

Summary
* Implements Faulty Temporal Memory Shim so that a TPRegion can used FaultyTM
* Update FulatyTM class according to the latest temporal_memory class in nupic
* Use FaultyTM in the discrete sequence prediction task (same as the one in the sequence memory paper)
* Fix the runtime error of when the percentage of cell death is high
* Remove connections when kill cells in FaultyTM
* Verify that changes to FaultyTM didn't affect the simulation of the Neuron paper


ywcui1990:
@subutai 
Prediction accuracy as a function of cell death. Learning is disabled after cell death in this experiment.
image

PR #450, opened 2016-01-29
Normalize Imbu scores for word-level models

BoltzmannBrain:
Normalize results to accommodate for multiple tokens in queries (https://jira.numenta.com/browse/RES-178).
@subutai merging to test Imbu.

### DIFF
--- a/htmresearch/frameworks/nlp/imbu.py
+++ b/htmresearch/frameworks/nlp/imbu.py
@@ -211,9 +211,6 @@ class ImbuModels(object):
     if loadPath:
       # User has explicitly specified a load path and expects a model to exist
       try:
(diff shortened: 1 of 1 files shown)


subutai:
@BoltzmannBrain This won't affect the order of the results, will it?

BoltzmannBrain:
Nope. The query() method in this class gets unsorted results from the classification model and the formatResults() method doesn't do any sorting, so the client (Imbu API) is responsible for sorting.

subutai:
Don't we sort results in classification model if sortResults=True?

BoltzmannBrain:
Yes, Imbu specifies sortResults=True. Leaving the results unsorted allows us to pass less information between the ImbuModels class and the Imbu API, speeding up the models.

PR #457, opened 2016-02-11
[PUB-64] Temporal Noise experiment with HTM and LSTM

ywcui1990:
@subutai 

This experiment test the performance of HTM and LSTM after temporal noise. 
First, we train HTM and LSTM on the high-order sequence prediction task until both models achieve perfect performance.

Second, we inject temporal noise. One element within the sequence will be replaced by a random element. The 2nd, 3rd, or 4th element of the sequence may be affect by such temporal noise. There is no further learning after the injection of the temporal noise.

We monitored the performance of HTM and LSTM on predicting the last element of the sequence after injected temporal noise.

The performance of HTM drops to 50% with the noise, which is expected because these sequences have 2 possible endings depending on the 1st element of the sequence. With the noise, TM lost the context of the very first element of the sequence, and will predict both endings.

It's somewhat unexpected that the performance of LSTM is even worse. It shows even more sensitivity to such temporal noise than HTM. I guess the performance might improve a bit if we train LSTM with noise. It may learn to ignore the noises.

image


ywcui1990:
@subutai I tested the performance of LSTM when it is trained with the temporal noise. It did improve a lot and matches HTM's performance

image

PR #458, opened 2016-02-11
Imbu 119 imbu models unittests

oxtopus:


tests/nlp/unit/imbu_models_tests.py
```diff
@@ -40,6 +47,94 @@ def setUp(self):
...
+    # Setup fake ImbuModels instance
+    imbu = ImbuModels(
+      cacheRoot="fake_cache_root",
+      dataPath=self.dataPath,
+      retina="en_associative",
+      apiKey=os.environ.get("CORTICAL_API_KEY")
+    )
+
+    # You must specify a known model type
+    self.assertRaises(Exception, imbu.createModel, "random model type")
```
BoltzmannBrain: I don't think this testing for the right exception. Should be `self.assertRaises(ValueError, imbu.createModel, modelType, None, None)`

tests/nlp/unit/imbu_models_tests.py
```diff
@@ -40,6 +47,94 @@ def setUp(self):
...
+
+    # Assert that you can create models of known types from scratch (i.e. with
+    # empty loadPath value)
+    for modelType in imbu.modelMappings:
+      # You _must_ specify loadPath and savePath.  Assert that attempts to not
+      # specify will result in failure
+      self.assertRaises(Exception, imbu.createModel, modelType)
```
BoltzmannBrain: Not testing for the right exception. Results in *** TypeError: createModel() takes at least 4 arguments (2 given)

tests/nlp/unit/imbu_models_tests.py
```diff
@@ -40,6 +47,94 @@ def setUp(self):
...
+      self.assertRaises(Exception, imbu.createModel, modelType, None)
```
BoltzmannBrain: Not testing for the right exception. Results in *** TypeError: createModel() takes at least 4 arguments (3 given)

tests/nlp/unit/imbu_models_tests.py
```diff
@@ -40,6 +47,94 @@ def setUp(self):
...
+      self.assertRaises(Exception, imbu.createModel, modelType, None, None)
```
BoltzmannBrain: Why not explicitly assert we get a ValueError?

BoltzmannBrain:
@oxtopus left some comments. 
@subutai it's okay to delete all uses of FastTemporalMemory (per [nupic #2998](https://github.com/numenta/nupic/pull/2998)) yes?

oxtopus:
Thanks.  I fixed up the exception types.  The FastTemporalMemory deletions are from https://github.com/numenta/nupic.research/pull/459 which I merged into here to validate that it causes the tests to pass.

BoltzmannBrain:
:+1:

subutai:
> @subutai it's okay to delete all uses of FastTemporalMemory (per nupic #2998) yes?

Yes, but per my JIRA detail, we can't just blindly delete. Where FTM is used we should replace it with just TM.  If we just delete it the scripts won't work anymore.

PR #481, opened 2016-03-07
NLP network params

BoltzmannBrain:
Adds a param file for Imbu to use TM.
@subutai I tried adding params from tm_knn_4k_retina that Yuwei has tuned a bit for NLP data to the tm2_knn_4k_retina params, but the following params were not added b/c they are unknown parameters for TM region type py.TMRegion: `maxSynapsesPerSegment`, `maxSegmentsPerCell`, `globalDecay`, `maxAge`, and `outputType`.


subutai:
:+1:

PR #494, opened 2016-03-16
Fix NLP cacheRoot bug

BoltzmannBrain:
@subutai the changes enable us to specify a cache directory for NLP models that use the Cio encoder; previously all would go to the default "CioCache/" dir.
@oxtopus this enables us to train models with a cache directory in a data volume that persists outside the container, and also fixes the simple UP model.

WIP, adding unit tests.


subutai:
@BoltzmannBrain Can we make sure it is optional, and defaults to the previous setting?

BoltzmannBrain:
@subutai it is. To check I covered the default setting with a unit test.
This is ready for review.
@oxtopus

htmresearch/frameworks/nlp/classify_network_api.py
```diff
@@ -57,6 +57,13 @@ def __init__(self,
       raise RuntimeError("Must specify a network config dict")
 
 
+  def getEncoder(self):
+    """
+    Returns the encoder instance for the model.
+    """
+    return self.encoder
```
subutai: This is not going to work properly after saving and loading. It should instead do:
```
return self.sensorRegion.getSelf().encoder
```
You can then get rid of self.encoder.

htmresearch/regions/LanguageSensor.py
```diff
@@ -104,6 +106,13 @@ def getSpec(cls):
       },
       "inputs":{},
       "parameters":{
+        "cacheRoot":{
```
subutai: I don't understand why you need this as a parameter. The encoder is being created outside of this region and set explicitly.

htmresearch/frameworks/nlp/classify_htm.py
```diff
@@ -50,6 +52,9 @@ def __init__(self,
     self.retina = retina
     self.apiKey = apiKey
     self.maxSparsity = maxSparsity
+    self.cacheRoot = cacheRoot
```
subutai: Why store cacheRoot here? It's stored in the encoder - best to get it from there rather than duplicate it. It also doesn't seem to be used anywhere other than initModel - you can just pass cacheRoot in to that method as a parameter.

htmresearch/regions/LanguageSensor.py
```diff
@@ -50,8 +50,10 @@ class LanguageSensor(PyRegion):
 
   def __init__(self,
                verbosity=0,
-               numCategories=1):
+               numCategories=1,
+               cacheRoot=None):
     """Create a node without an encoder or datasource."""
+    self.cacheRoot = cacheRoot
```
subutai: Why have a member variable self.cacheRoot here?

subutai:
@BoltzmannBrain Does this really solve the underlying problem? If you train with a certain cacheDir and then load the network on another machine with a different path for cacheDir, I don't see a way to change the cacheDir dynamically in this implementation.  In addition I don't think the tests cover this scenario.

All you need to do is implement a setCacheDir() method in cio_encoder.py and then, once you load the model, do model.getEncoder().setCacheDir(NEWPATH).

oxtopus:
@subutai The problem, as I understand it, is the cortipy _client_, itself, is being serialized along with the encoder.  When the models are trained on one machine, saved, and then loaded on another, the client will have the original parameters, which may be different.

Personally, I don't see any reason why the client needs to be serialized and deserialized.  It should be reinitialized in the new environment.  For example, this is exactly the case described in the docstring for https://github.com/BoltzmannBrain/nupic.research/blob/imbu_simpleUP/htmresearch/encoders/cio_encoder.py#L124-L134

Furthermore, I don't think we need to be setting the cache dir at all, and instead, rely on the client to handle the default location.  When it comes to caching, moving from one environment to another should gracefully handle the absence of prior cache values.  For example, if you train it locally, you will have built up a cache.  Then, if you transfer the trained models into an environment in which there is no cache, the client should transparently restore the cache on-demand.

Should we want to preserve the cache, then we can handle that with volume mappings for the container.

oxtopus:
re:

> All you need to do is implement a setCacheDir() method in cio_encoder.py and then, once you load the model, do model.getEncoder().setCacheDir(NEWPATH).

This!

oxtopus:
I also just realized that the client is constructed twice: https://github.com/BoltzmannBrain/nupic.research/blob/imbu_simpleUP/htmresearch/encoders/cio_encoder.py#L83-L89

If you remove the assignment of `self.cacheDir` and the second assignment of `self.client`, defaults will be used automatically in the new environment where the encoder is loaded.

subutai:
> I also just realized that the client is constructed twice:

Yikes, good catch. That is in the current codebase also, before the PR.

subutai:
> This!

@oxtopus Agreed :smile:   

With regards to whether we serialize or deserialize the client, I think it is fine to have CioEncoder not save the client when pickled, and instead reinitialize it each time it is unpickled.

oxtopus:
>  I think it is fine to have CioEncoder not save the client when pickled, and instead reinitialize it each time it is unpickled.

That is basically what already happens, or is supposed to happen.  With https://github.com/BoltzmannBrain/nupic.research/blob/imbu_simpleUP/htmresearch/encoders/cio_encoder.py#L86-L89 removed, that should happen automatically.

BoltzmannBrain:
> Furthermore, I don't think we need to be setting the cache dir at all, and instead, rely on the client to handle the default location. When it comes to caching, moving from one environment to another should gracefully handle the absence of prior cache values. For example, if you train it locally, you will have built up a cache. Then, if you transfer the trained models into an environment in which there is no cache, the client should transparently restore the cache on-demand.

> Should we want to preserve the cache, then we can handle that with volume mappings for the container.

:+1: :+1: this is a good plan, and avoids the common-cache mess I was trying make work.
Thank you both for the detailed notes and review, I'll followup with appropriate changes later when I have time.
@oxtopus this plan also influence https://github.com/NumentaCorp/grok-projects/pull/242, which I'll fix accordingly.

htmresearch/regions/LanguageSensor.py
BoltzmannBrain: B/c it wasn't clear the cacheDir param for the encoder was being set properly, so this seemed to fix it, but this is a bad fix. The region shouldn't care about the cache; the cache should be set by the encoder and handled by the Cio client.

htmresearch/frameworks/nlp/classify_htm.py
BoltzmannBrain: You're right, thanks, I set this before adding the getEncoder() method.

htmresearch/frameworks/nlp/classify_network_api.py
BoltzmannBrain: Thank you.

BoltzmannBrain:
@oxtopus please review

BoltzmannBrain:
@subutai please take a look; I added the int test we discussed.

tests/imbu/integration/imbu_model_lifecycle_tests.py
```diff
@@ -0,0 +1,110 @@
...
+    originalModel = imbu.createModel(modelType,
+                                     loadPath="",
+                                     savePath=checkpointLocation,
+                                     networkConfigName=networkConfigName)
+
+    # Physically move the cache to a new directory; we copy it b/c moving it
+    # would break the test cleanup method
+    encoder = originalModel.getEncoder()
+    newCacheDir = os.path.join(imbuTempDir, "new_test_cache")
+    shutil.copytree(originalCacheDir, newCacheDir)
```
subutai: You might want to delete the original cache dir. before doing loading the new model.

tests/imbu/integration/imbu_model_lifecycle_tests.py
```diff
@@ -0,0 +1,110 @@
...
+
+class TestImbuLifecycle(unittest.TestCase):
+
+  def setUp(self):
+    self.dataPath = os.path.join(TEST_DATA_DIR, "sample_reviews_subset.csv")
+
+
+  def _createTempDir(self):
+    tmpDir = tempfile.mkdtemp()
+    self.addCleanup(shutil.rmtree, tmpDir)
```
subutai: Don't need this cleanup if you explicitly delete below.

BoltzmannBrain: The new cache directory gets created in the call to `imbu.createModel()`. If something in that call breaks the test, the cache directory gets deleted in the test clean up. Instead, if I explicitly delete it later, the cache directory would remain if `imbu.createModel()` breaks the test.

subutai: Ah ok, makes sense.

tests/imbu/integration/imbu_model_lifecycle_tests.py
```diff
@@ -0,0 +1,110 @@
...
+      dataPath=self.dataPath,
+      retina="en_associative",
+      apiKey=os.environ.get("CORTICAL_API_KEY")
+    )
+
+    # Train with standard cache location, save
+    modelType = "HTMNetwork"
+    networkConfigName = "imbu_sensor_knn.json"
+    checkpointLocation = os.path.join(imbuTempDir, "checkpoint")
+    originalModel = imbu.createModel(modelType,
```
subutai: Where does the model get trained?

BoltzmannBrain: In `imbu.createModel()`: https://github.com/numenta/nupic.research/blob/master/htmresearch/frameworks/nlp/imbu.py#L261

htmresearch/encoders/cio_encoder.py
```diff
@@ -135,11 +138,11 @@ def __setstate__(self, state):
     self.__dict__ = state
 
 
-  def _setDimensions(self, retina, scalingFactor):
+  def _setDimensions(self, scalingFactor):
     if scalingFactor <= 0 or scalingFactor > 1:
       raise ValueError("Retina can only be scaled by values between 0 and 1.")
 
-    retinaDim = RETINA_SIZES[retina]["width"]
+    retinaDim = RETINA_SIZES[self.retina]["width"]
```
subutai: if self.retina is only ever used here (called when initialized) then do we really need a self.retina member variable?

subutai:
Left a few minor comments - mainly not sure the integration test is actually training the model - ???

htmresearch/encoders/cio_encoder.py
BoltzmannBrain: It's used when reinstantiating the Cio client in the cacheDir setter.

tests/imbu/integration/imbu_model_lifecycle_tests.py
subutai: OK. What data file is it using? You might want a small one for the purposes of this test.

subutai:
OK. :+1:

tests/imbu/integration/imbu_model_lifecycle_tests.py
BoltzmannBrain: It's using a small subset of the sample reviews data; only two reviews.

BoltzmannBrain:
Thank you @subutai

PR #496, opened 2016-03-18
Add unit tests for SDR classifier

ywcui1990:
@subutai 

I have added unit tests for the SDR classifier. These are the same set of tests as in https://github.com/numenta/nupic/blob/master/tests/unit/nupic/algorithms/cla_classifier_test.py

Except
* The expected results of the tests are adapted for the SDR classifier
* the WriteRead test has not been included, since we don't have the capnp proto for the new classifier yet.
* One obsolete tests and one disabled tests were not included

@subutai @scottpurdy 
The SDR classifier are ready for code review now. Should we do it here in nupic.research, or should I create a separate PR for nupic?

### DIFF
--- a/htmresearch/algorithms/sdr_classifier.py
+++ b/htmresearch/algorithms/sdr_classifier.py
@@ -34,6 +34,7 @@ g_debugPrefix = "SDRClassifier"
 def _pFormatArray(array_, fmt="%.2f"):
   """Return a string with pretty-print of a numpy array using the given format
   for each element"""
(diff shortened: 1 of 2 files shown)


subutai:
@ywcui1990 I think we should do the PR in NuPIC. 

BTW, I notice that the comments need to be cleaned up - there are a lot of references to "neural networks".   The comments should be focused on SDRs and HTMs. You can keep the existing references to refer to the actual training method.

PR #502, opened 2016-03-24
Junit tests

subutai:


subutai:
@BoltzmannBrain RES-197 and some misc changes.

projects/nlp/junit_test3.py
```diff
@@ -117,7 +130,7 @@ def run(args):
                       default="junit3_checkpoints",
                       help="Model(s) will be saved in this directory.")
   parser.add_argument("--retina",
-                      default="en_associative_64_univ",
+                      default="en_synonymous",
```
BoltzmannBrain: Why for this test only?

htmresearch/support/text_preprocess.py
```diff
@@ -249,7 +249,7 @@ def _filterStuff(self,
 
     if removeStrings:
       for removal in removeStrings:
-        textString = textString.replace(removal, "")
+        textString = textString.replace(removal.lower(), "")
```
BoltzmannBrain: Why?
```
>>> text = "Hello world!"
>>> text.replace("Hello", "hi")
'hi world!'
>>> text.replace("Hello".lower(), "hi")
'Hello world!'
```

BoltzmannBrain:
@subutai left a couple questions, otherwise :+1:

htmresearch/support/text_preprocess.py
subutai: Good catch. Originally I did it thinking we need it this way since tokenize() lower cases everything. But now I see the lower casing happens after string removal.

projects/nlp/junit_test3.py
subutai: Yes, I should do it for the others too.

PR #503, opened 2016-03-25
RES-185: NLP model tests

BoltzmannBrain:
Formalizing (automating) the manual API tests [here](https://www.evernote.com/l/AAwmQ9c9-QxFBIq7z7wUzMKr7BMomvQLgY8)


BoltzmannBrain:
@subutai please take a look.
I've confirmed the results are consistent for all the models and tests, but for one: running the simple_labels test with docfp gives me 63.8%, not the expected 99.7%. And when I run CioDocumentFingerprint, I get 100% (both here and on master). I'm running w/ `python simple_labels.py --dataPath FILE -m docfp --retina en_associative_64_univ --numLabels 10`.

subutai:
> I've confirmed the results are consistent for all the models and tests, but for one: running the simple_labels test with docfp gives me 63.8%, not the expected 99.7%.

Not sure why this dropped. On master (before your change) I get 99.7% for simple_labels.py using the same retina. Moving the files should cause accuracy to drop.

BoltzmannBrain:
@subutai I moved the files back so now we can see the diff, but it's not much clearer b/c I moved the common functions into a single file.

Moving the files now we run docfp on simple_queries with `python simple_labels.py -m docfp --dataPath FILE -c data/network_configs/sensor_knn.json`. This gets me 76.3%. When I specify `--retina en_associative_64_univ` I get 63.8%.

subutai:
@BoltzmannBrain Hmm, not sure what to say except that this needs to be debugged!  This refactoring shouldn't change the results, since you didn't make any algorithmic changes.

BoltzmannBrain:
@subutai please review. The discrepancy was b/c I had set `k=3` for docfp, when it should be `1` 😑

BoltzmannBrain:
@subutai please review

subutai:
@BoltzmannBrain Whoah - why are we adding all these lines of code to hello_classification_model.py?  They are unnecessary and making the file much longer and complex than it needs to be.  This is not a generic file - it is literally supposed to be "hello world".  Can't we just keep it the way it was?

projects/nlp/hello_classification_model.py
```diff
@@ -23,44 +23,55 @@
...
 from htmresearch.frameworks.nlp.model_factory import (
   createModel, getNetworkConfig)
+from htmresearch.support.csv_helper import mapLabelRefs, readCSV
+
+
+wrapper = TextWrapper(width=80)
+
+
+
+def getData(dataPath):
```
subutai: For the "hello" example file I would prefer to keep the data explicit within the file rather than have a level of indirection and read in another file. Much simpler to understand (you don't need to open and understand a file format and the usage of mapLabelRefs).  Note that you are not saving any lines of code by adding this getData method. It is also harder to understand the explanation comments above because they are abstracted from the data.

projects/nlp/hello_classification_model.py
```diff
@@ -73,89 +84,129 @@ def instantiateModel(args):
...
-  print "==========================Classifying sample text================"
-  numCorrect = 0
-  for docId, doc in enumerate(testData):
-    document = doc[0]
-    desiredLabels = doc[1]
+  print "==============Testing {} on sample text=============".format(modelName)
+  if verbosity > 0:
     print
-    print "Document=", document,", desired label: ",desiredLabels
+    printTemplate = PrettyTable(
```
subutai: All these lines are getting in the way of understanding the core API. We don't need them in this example.

projects/nlp/model_api_tests/hello_classification_model.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python
```
subutai: Why do you need this file in model_api_tests directory? I thought we were going to use simple_labels and read the data from a file?

subutai:
@BoltzmannBrain I feel like this PR is supposed to just add model tests. However it is also significantly changing the example code, which is a different effort. It is actually making the simple examples harder to understand. Can't we just leave the existing example files exactly as-is and just add additional test files in a separate directory?  I'm not sure why we are mixing the two.

BoltzmannBrain:
Thanks @subutai, I was also concerned with the add'l overhead in the hello classification script affecting readability. I'll make it more simple, like a true hello world.

projects/nlp/model_api_tests/hello_classification_model.py
BoltzmannBrain: Sorry forgot to remove it

projects/nlp/hello_classification_model.py
BoltzmannBrain: Because the same "hello classification" data is read in with the simple_labels.py script I did the same here as not to duplicate the data. Sure, I can switch it, we want people to easily grasp the classification model API with this example, not confuse them.

subutai:
> I'll make it more simple, like a true hello world.

Can we just keep it (and simple_labels) exactly the way it was? Just add the additional files you need for the test itself. If we want to change the example files, that should be a different PR and a different effort.

BoltzmannBrain:
@subutai should be all set now.

htmresearch/frameworks/nlp/classify_document_fingerprint.py
```diff
@@ -163,7 +162,7 @@ def inferToken(self, token, reset=0, returnDetailedResults=False,
       document = " ".join(self.currentDocument)
       sensor = self.sensorRegion.getSelf()
       sensor.addDataToQueue(token=document, categoryList=[None],
-                            sequenceId=-1, reset=0)
+                            sequenceId=-1, reset=reset)
```
oxtopus: It would be nice if `reset` had a more descriptive name.  It's not well documented in the docstring, and it comes up in various places (like this line) and it's not clear *what* `reset` is supposed to do.  Also, since it's being used as a boolean, it would be either `True` or `False` and not integer `1` or `0`.

htmresearch/support/nlp_model_test_helpers.py
```diff
@@ -0,0 +1,242 @@
...
+  },
+}
+
+wrapper = TextWrapper(width=80)
+
+
+
+def executeModelLifecycle(args, trainingData, labelRefs):
+  """ Execute model lifecycle: create a model, train it, save it, reload it.
+  Returns both the original and new models.
```
oxtopus: Perhaps include detail that the return value is a tuple comprising both models.

htmresearch/support/nlp_model_test_helpers.py
```diff
@@ -0,0 +1,242 @@
...
+    "HTM_sensor_tm_knn": 0.0,
+    "keywords": 16.9,
+  },
+}
+
+wrapper = TextWrapper(width=80)
+
+
+
+def executeModelLifecycle(args, trainingData, labelRefs):
```
oxtopus: What is `args`?  I see that it is passed along to `instantiateModel()`, but there's no description of its structure or expected values.

htmresearch/support/nlp_model_test_helpers.py
```diff
@@ -0,0 +1,242 @@
...
+
+# Network models use 4k retina.
+htmConfigs = [
+  ("HTM_sensor_knn", "../data/network_configs/sensor_knn.json"),
+  ("HTM_sensor_simple_tp_knn", "../data/network_configs/sensor_simple_tp_knn.json"),
+  ("HTM_sensor_tm_knn", "../data/network_configs/sensor_tm_knn.json"),
+]
+
+# Some values of k we know work well.
+kValues = { "keywords": 21, "docfp": 1}
```
oxtopus: Consider capitalizing as `K_VALUES`.

oxtopus: Additionally, the rest of the global module-level variables should follow suit.

htmresearch/support/nlp_model_test_helpers.py
```diff
@@ -0,0 +1,242 @@
...
+  Test the given model on testData, print out and return accuracy percentage.
+
+  Accuracy is calculated as follows. Each token in a document votes for a single
+  category; it's possible for a token to contribute no votes. The document is
+  classified with the category that received the most votes. Note that it is
+  possible for a document to receive no votes, in which case it is counted as a
+  misclassification.
+  """
+  modelName = repr(model).split()[0].split(".")[-1]
+  print
```
oxtopus: This output is not within a `if verbosity` block.

htmresearch/support/nlp_model_test_helpers.py
```diff
@@ -0,0 +1,242 @@
...
+def assertResults(testName, accuracies):
+  """ Assert the new acuracies against the current values.
+  @param testName (str) One of the nlpModelAccuracies keys.
+  @param accuracies (dict) Keys are model names, values are accuracy percents.
+  """
+  try:
+    currentAccuracies = nlpModelAccuracies[testName]
+  except KeyError as e:
+    print "No accuracy values for test '{}'".format(testName)
+    raise e
```
oxtopus: Need only be `raise`.

projects/nlp/model_api_tests/simple_labels.py
```diff
@@ -0,0 +1,195 @@
...
+  else:
+    modelNames = [args.modelName]
+    runningAllModels = False
+
+  accuracies = {}
+  for name in modelNames:
+    # Setup args
+    args.modelName = name
+    args.modelDir = os.path.join(args.experimentName, name)
+    if runningAllModels and name == "htm":
```
oxtopus: Consider simplifying as:

```python
if name == "htm":
  if runningAllModels:
    ...
  else:
    ...
```

projects/nlp/model_api_tests/simple_labels.py
```diff
@@ -0,0 +1,195 @@
...
+      split = int(len(dataset) * args.split)
+      trainingData = dataset[:split]
+    else:
+      trainingData = dataset
+
+    # Create a model, train it, save it, reload it
+    _, model = executeModelLifecycle(args, trainingData, labelRefs)
+
+    # Test the model
+    accuracies.update({name:testModel(model,
```
oxtopus: This might be easier to read as:

```python
accuracies["name"] = testModel(...)
```

htmresearch/support/nlp_model_test_helpers.py
```diff
@@ -0,0 +1,242 @@
...
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+helpStr = """
+Methods and data for running NLP model API tests. The intent here is
+to ensure and changes to the models does not decrease their classification
+accuracies (see nlpModelAccuracies below). Three tests are supported:
```
subutai: "ensure that changes to the models do not"

projects/nlp/simple_query.py
```diff
@@ -1,219 +0,0 @@
-#!/usr/bin/env python
```
subutai: Should we retain simple_query as an example script in projects/nlp?

subutai:
I'm ok with these changes. I did not go through code in detail, but looks like @oxtopus did.

htmresearch/frameworks/nlp/classify_document_fingerprint.py
BoltzmannBrain: Thanks, I moved the docfp changes to a new PR (#531).

BoltzmannBrain:
Thank you @oxtopus and @subutai for the reviews.

PR #519, opened 2016-04-05
Imbu fixes -- query with multiple words, dataset name mappings

BoltzmannBrain:


BoltzmannBrain:
@subutai merging so we can deploy imbu

PR #546, opened 2016-05-11
Add support for TemporalMemoryCPP in TMRegion

subutai:
Fix bug in getConstructorArguments in temporal_memory_factory
Use the proper accessor methods to get active/predictive cells
Validate with hello_classification_model

### DIFF
--- a/htmresearch/algorithms/temporal_memory_factory.py
+++ b/htmresearch/algorithms/temporal_memory_factory.py
@@ -77,14 +77,25 @@ def createModel(modelName, **kwargs):
 
 def getConstructorArguments(modelName):
   """
-  Return a list of strings corresponding to constructor arguments for the
+  Return constructor arguments and associated default values for the
   given model type.
 
   @param modelName (str)  A supported temporal memory type
 
+  @return argNames (list of str) a list of strings corresponding to constructor
+                                 arguments for the given model type. The first
+                                 item is usually 'self'.
+  @return defaults (list)        a list of default values for each argument
+                                 (excluding self)
   """
 
   if modelName not in TemporalMemoryTypes.getTypes():
     raise RuntimeError("Unknown model type: " + modelName)
 
-  return inspect.getargspec(getattr(TemporalMemoryTypes, modelName).__init__)[0]
+  return (
+    inspect.getargspec(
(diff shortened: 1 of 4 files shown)


subutai:
@oxtopus Can you validate that I did the right thing in getConstructorArguments? I believe the previous version didn't do the right thing (it was a copy of createModel)

oxtopus:
Hmm.  yes, the status quo `getConstructorArguments()` looks suspiciously broken.  Your fix looks good, assuming you're _only_ concerned with args and not the full function signature.  Since `inspect.getargspec()` returns a named tuple, you may want to access it by name instead of index:

```
return inspect.getargspec(getattr(TemporalMemoryTypes, modelName).__init__).args
```

subutai:
@oxtopus Thanks. The comments string said it was to return a list of argument names only, but I think additional information would be useful. I don't think anyone else is using it, so I can make the change easily. Does this look ok?

oxtopus:
Something else to consider, since `__init__()` is an instance method, its argspec includes `self`.  So you may need to strip that out.

subutai:
@oxtopus Yeah, I had put that in the comments, but I think it is cleaner to strip it out so just pushed that. Thanks.

htmresearch/algorithms/temporal_memory_factory.py
```diff
@@ -77,14 +77,25 @@ def createModel(modelName, **kwargs):
...
+                                 item is usually 'self'.
+  @return defaults (list)        a list of default values for each argument
+                                 (excluding self)
   """
 
   if modelName not in TemporalMemoryTypes.getTypes():
     raise RuntimeError("Unknown model type: " + modelName)
 
-  return inspect.getargspec(getattr(TemporalMemoryTypes, modelName).__init__)[0]
+  return (
```
oxtopus: Better to not repeat the call to `inspect.getargspec()`.  Instead, do something like:

```python
argspec = inspect.getargspec(...)
return (argspec.args, argspec.defaults)
```

or:

```python
from operator import attrgetter
...
return attrgetter("args", "defaults")(inspect.getargspec(...))
```

oxtopus:
👍  pending deduping.

PR #549, opened 2016-05-11
Added ability chose default output in TMRegion

subutai:
Added ability chose default output in TMRegion - this allows you to easily control whether active or predictedActive cells get stored in the classifier.

Also added separate activeCells output you can always get all outputs.

Made print statements a bit friendlier in classify_network_api


subutai:
@BoltzmannBrain Let me know if you see any issues. Note, I don't think this is backward compatible but not sure it's worth the effort to make it so.

htmresearch/regions/TMRegion.py
```diff
@@ -374,11 +393,20 @@ def compute(self, inputs, outputs):
     self.previouslyPredictedCells[self._tm.getPredictiveCells()] = 1
 
     # Copy numpy values into the various outputs
-    # outputs['bottomUpOut'][:] = self.activeState
-    outputs['bottomUpOut'][:] = predictedActiveCells
+    outputs['activeCells'][:] = self.activeState
```
BoltzmannBrain: I know these were single quotes initially, but it'd be good to switch to double quotes here and below.

htmresearch/regions/TMRegion.py
```diff
@@ -117,22 +116,32 @@ def getSpec(cls):
         ),
         outputs=dict(
             bottomUpOut=dict(
-              description="""The primary output of the temporal memory.""",
+              description="""The default output of TMRegion. By default this
+                   outputs the active cells. You can change this dynamically
+                   using the defaultOutputType parameter.
+                   """,
```
BoltzmannBrain: Format is inconsistent with other description strings.

BoltzmannBrain:
@subutai I left a couple nitpicks but otherwise 👍 
If we run into issues w/ backwards compatibility it'll be a good forcing function to clean up the network configs dir.

subutai:
> If we run into issues w/ backwards compatibility it'll be a good forcing function to clean up the network configs dir.

@BoltzmannBrain Thanks. I believe the JSON files will continue to work. It's just that previously saved checkpoints of networks with TMRegion won't load anymore.

htmresearch/regions/TMRegion.py
vitaly-krugl: @subutai: an additional nitpick on top of Alex's - the text wrapped thusly doesn't print nicely (it even rhymes! :) )

```
In [3]: s = """abc
             def
             geh
             """

In [4]: print s
abc
             def
             geh
```

subutai: @vitaly-krugl Thanks - I think my recent push fixed it.  However I do now see there are other places in the file where the same mistake is made. I can take care of that in a later PR. Sorry, my reply doesn't rhyme! 😄

vitaly-krugl: No problem, like I said - just a nitpick. Better do something about the rhyming, though!

PR #563, opened 2016-06-16
Res 224

subutai:
Empty L2 layer implementation
Simple L4->L2 network construction


subutai:
@scottpurdy This is a simple network factory, like we discussed.

scottpurdy:
Looks good so far!

subutai:
@scottpurdy Perhaps you can do a quick review as a sanity check?

htmresearch/frameworks/layers/laminar_network.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env python
```
scottpurdy: nit: not need for this line

htmresearch/regions/ExtendedTMRegion.py
```diff
@@ -0,0 +1,449 @@
...
+    the constructor parameters for each class, and send it to that constructor.
+    """
+    if self._tm is None:
+      # Create dict of arguments we will pass to the temporal memory class
+      args = copy.deepcopy(self.__dict__)
+      args["columnDimensions"] = (self.columnCount,)
+      args.pop("columnCount")
+
+      # Create the TM instance
+      self._tm = createModel("general", **args)
```
scottpurdy: Instead of checking whether `self._tm` is set, why not just create it in the `__init__` function?

scottpurdy: I also think it's a little wacky to pass the instances `__dict__` members as keyword arguments. This is problematic since adding a new instance variable requires a change in here to remove it before passing `args` to the `createModel` function.

scottpurdy:
Aside from nit picks the only real concern I have is the region creation and whether to initialize things in `__init__` or `initialize`.

htmresearch/regions/ExtendedTMRegion.py
subutai: > I also think it's a little wacky to pass the instances __dict__ members as keyword arguments. This is problematic since adding a new instance variable requires a change in here to remove it before passing args to the createModel function.

This is copy-paste mistake from some other code. I actually don't need to pop anything.  [wh]acky but requires no maintenance!

scottpurdy: Oh I wasn't familiar with this. And @natoromano just ran into a problem in SDRClassifierRegion where it was clobbering the deserialized instance in `initialize` so I was trying to avoid that issue. I'll think about how to make it clear to developers how to do this properly since right now it really isn't obvious example how to handle `__init__`/`initialize`.

PR #567, opened 2016-07-06
Adds test for Extended Temporal Memory

natoromano:
This PR adds a test suite for Extended Temporal Memory (ETM), formerly known as General Temporal Memory.

In particular, it adds three series of tests:

- **extended_temporal_memory_test.py** is a series of unit tests, testing very basic low-level aspects of ETM. Most of them are drawn from the regular Temporal Memory (TM) tests, and some are added to test specific aspects of ETM. In particular, 3 unit tests were added. The serialization test is skipped for now as no serialization is available for ETM.

- **regular_etm_test.py** is a series of algorithmic tests, that is porting TM tests to ETM, and thus are only testing the sequence memory aspects. The only difference with the TM tests is that they have been adapted to use typical default parameters, rather than made-up ones. The tests where only slightly changed to adapt to this change (more iterations were needed for example).

- **extensive_etm_test.py** is a completely new series of algorithmic tests, which are testing the specific aspects of ETM, i.e. learning through external input, learning through apical input, learning on one cell, etc. Since all tests are new, this is probably the most interesting to review.

In its current state, all test pass, except one, from the last series. It is likely to be showing a slight problem when using learnOnOneCell, which I think I figured out and we can discuss.

Ready for review @subutai.


tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1274 @@
...
+    predictedActiveB = self.tm.mmGetTracePredictedActiveCells().data[-1]
+    # check that the two representations are the same
+    self.assertEqual(predictedActiveA, predictedActiveB)
+
+
+  def testE14(self):
+    """learnOnOneCell with motor command should not impair behavior.
+
+    Using learnOnOneCell, does the same test as E6, using the same parameters.
+    """
```
natoromano: This is the test that fails. Whereas we expect the ETM to learn this simple pattern, a few columns (0 to 5) burst at each step. This is because of the winning cell mechanism, which has a slight flaw when using learnOnOneCell: since each synapse needs two "increments" to become connected, and the least used cell is chosen, for each column in the first pattern of the sequence, all cells will have to be selected once before going back to the first one and connecting the synapse. The consequence of this is that the sequence has to be fed exactly (#columns + 1) times for the test to pass.

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1274 @@
...
+    proximalInputA = self.sequenceMachine.generateFromNumbers(numbers)
+    numbers = self.sequenceMachine.generateNumbers(1, 100)
+    externalInputA = self.sequenceMachine.generateFromNumbers(numbers)
+
+    for _ in xrange(2):
+      self.feedTM(proximalInputA, activeExternalCellsSequence=externalInputA)
+
+    self._testTM(proximalInputA, activeExternalCellsSequence=externalInputA)
+
+    self.assertAllActiveWerePredicted()
```
natoromano: This is the test that fails. Whereas we expect the ETM to learn this simple pattern, a few columns (0 to 5) burst at each step. This is because of the winning cell mechanism, which has a slight flaw when using learnOnOneCell: since each synapse needs two "increments" to become connected, and the least used cell is chosen, for each column in the first pattern of the sequence, all cells will have to be selected once before going back to the first one and connecting the synapse. The consequence of this is that the sequence has to be fed exactly (#columns + 1) times for the test to pass.

tests/extended_temporal_memory/regular_etm_test.py
```diff
@@ -0,0 +1,676 @@
+#!/usr/bin/env python
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2014, Numenta, Inc.  Unless you have an agreement
```
subutai: 2016

subutai:
The filenames "extended vs extensive" are a bit confusing. Perhaps rename `extended_temporal_memory_test.py` to `etm_unit_test.py`?

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+
+      columnDimensions: (2048,)
+      cellsPerColumn: 32
+      initialPermanence: 0.5
+      connectedPermanence: 0.6
+      minThreshold: 25
+      maxNewSynapseCount: 30
+      permanenceIncrement: 0.1
+      permanenceDecrement: 0.02
+      predictedSegmentDecrement: 0.08
```
subutai: Typically we use a much smaller value for predictedSegmentDecrement, about 10% or 5% of the increment

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+  each columns between subsequent resets.
+
+  E14: "learnOnOneCell" with motor command should not impair behavior.
+
+  ============================================================================
+                        Learning with apical input
+  ============================================================================
+
+  These tests (labeled "A"0, simulate sequences from proximal input as well as
+  sequences of feedback from higher regions. It tests some basic properties
```
subutai: "as well as feedback from higher regions connected to apical dendrites."

tests/extended_temporal_memory/extensive_etm_test.py
subutai: Hmm, is this because there are some shared columns between the various patterns?

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    externalInputA = self.sequenceMachine.generateFromNumbers(numbers)
+
+    numbers = self.sequenceMachine.generateNumbers(1, 100)
+    externalInputB = self.sequenceMachine.generateFromNumbers(numbers)
+
+    for _ in xrange(2):
+      self.feedTM(proximalInputA, activeExternalCellsSequence=externalInputA)
+
+    self._testTM(proximalInputA, activeExternalCellsSequence=externalInputB)
+    self.assertAlmostAllActiveWereUnpredicted()
```
subutai: Why "almost all", why not all?

tests/extended_temporal_memory/extensive_etm_test.py
natoromano: @subutai yes exactly. If there is just one collision between two patterns in the sequence (and, to be more specific, between the first one and another one), the sequence will have to be fed (#columns + 1) times instead of 2.

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    # test on X B C D E
+    self._testTM(testProximalInput,
+                 activeExternalCellsSequence=externalInputA)
+
+    unpredictedActiveColumns = self.tm.mmGetTraceUnpredictedActiveColumns().data
+    self.assertGreaterEqual(unpredictedActiveColumns, min(self.w))
+
+
+  def testE11(self):
+    """Repeated motor command copy as external input."""
```
subutai: What does this comment mean? Can you be more explicit?

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    # few extra predictions
+    predictedInactiveColumns = self.tm.mmGetTracePredictedInactiveColumns().data
+    self.assertLessEqual(len(predictedInactiveColumns[-1]), 3)
+
+
+  def testA3(self):
+    """Basic feedback disambiguation, test with correct feedback.
+
+    Train on ABCDE with F1, XBCDE with F2.
+    Test with BCDF. With feedback, one pattern is expected.
```
subutai: Do you mean "XBCDF" with F2?

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    self._testTM(testProximalInput, activeApicalCellsSequence=feedbackB)
+    # burst at last step
+    unpredictedActiveColumns = self.tm.mmGetTraceUnpredictedActiveColumns().data
+    self.assertGreaterEqual(len(unpredictedActiveColumns[-1]), min(self.w))
+
+
+  def testA5(self):
+    """Robustness to temporal noise with feedback, test without feedback.
+
+    Train on ABCDE with F1, XBCDE with F2.
```
subutai: XBCDF

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    # many extra predictions
+    predictedInactiveColumns = self.tm.mmGetTracePredictedInactiveColumns().data
+    self.assertGreaterEqual(len(predictedInactiveColumns[-1]), min(self.w))
+
+
+  def testA6(self):
+    """Robustness to temporal noise with feedback, test with incorrect
+    feedback.
+
+    Train on ABCDE with F1, XBCDE with F2.
```
subutai: XBCDF

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    self._testTM(testProximalInput, activeApicalCellsSequence=feedbackA)
+    # burst at last step
+    unpredictedActiveColumns = self.tm.mmGetTraceUnpredictedActiveColumns().data
+    self.assertGreaterEqual(len(unpredictedActiveColumns[-1]), min(self.w))
+
+
+  def testA7(self):
+    """Robustness to temporal noise with feedback, test without feedback.
+
+    Train on ABCDE with F1, XBCDE with F2.
```
subutai: XBCDF

tests/extended_temporal_memory/extensive_etm_test.py
natoromano: @subutai the randomly generated sequence of external inputs happen to have the same pattern at some index (36 to be precise), and thus the prediction is correct at the next step. I figured this was less of a hack as manually changing this pattern, as we would expect this to happen in real settings as well.

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    # burst at last step
+    unpredictedActiveColumns = self.tm.mmGetTraceUnpredictedActiveColumns().data
+    self.assertGreaterEqual(len(unpredictedActiveColumns[-1]), min(self.w))
+
+
+  def testA9(self):
+    """Robustness to temporal noise with feedback, test with correct feedback.
+
+    Train on ABCDE with F1, XBCDE with F2.
+    Test with AZCDF (one step is corrupted). With feedback F2, no bursting,
```
subutai: XBCDF. Confused by this one. You are starting with A but feedback is F2?

tests/extended_temporal_memory/extensive_etm_test.py
natoromano: Yes.

tests/extended_temporal_memory/extensive_etm_test.py
natoromano: Wow this is weird indeed. I'm looking into it.

tests/extended_temporal_memory/extensive_etm_test.py
```diff
@@ -0,0 +1,1340 @@
...
+    # burst at last step
+    unpredictedActiveColumns = self.tm.mmGetTraceUnpredictedActiveColumns().data
+    self.assertGreaterEqual(len(unpredictedActiveColumns[-1]), min(self.w))
+
+
+  def testO4(self):
+    """Joint external / apical inputs, with correct feedback.
+
+    Train on ABCDE / PQRST with F1, XBCDEF / PQRST with F2.
+    Test with BCDE / QRST. With feedback F1, no burst nor extra (feedback
```
subutai: This is a good test, but do you want to have a unique ending for the first sequence?

subutai:
@natoromano Overall this looks excellent - very clean code and a good set of tests. I would like to review some of them with you tomorrow. We might want more extensive E-series of tests. It would also be good to include some tests with unions.

tests/extended_temporal_memory/extensive_etm_test.py
natoromano: @subutai What do you mean? (btw again it should be XBCDF, sorry about those)

natoromano:
Merging, #572 will have to be addressed.

PR #573, opened 2016-07-16
Additional extensive tests for ETM

### LINKED ISSUE #572: Add a few extra tests to ETM extensive tests
Add the following tests to ETM tests:

- [x] test that ambiguous sequences predict union of representations (different patterns, same pattern but different representations)
- [x] test that union of feedback predicts union of representations
- [x] should also work with learnOnOneCell enabled
- [x] test that disambiguation on the feedback leads to disambiguation on the predictions
- [x] add external input on the learnOnOneCell test
- [x] loosen failing test, add test with #columns + 1 iterations
- [x] add test that exhibit #565

natoromano:
Fixes #572, adding tests for the extended temporal memory, as discussed.

@subutai


subutai:
👍

PR #580, opened 2016-07-26
Add a numentaTM detector that doesn't use the network API

mrcslws:
@BoltzmannBrain


BoltzmannBrain:
Awesome, thank you. I'm going to go ahead and merge so I can use it.

PR #587, opened 2016-08-04
RES-281 Online agglomerative clustering

marionleborgne:
*RES-275*
I updated the signal generator to run experiments with:
* A sine wave (different frequency for each category)
* A periodic binary signal (different frequency for each category)
* A triangular signal (different sampling rate for each category)
unit_data

You can find the results of the experiments in `nupic.research/projects/classification/sensor/artificial_data/results/seq_classification_100_reps.csv`

* To run the experiments: `nupic.research/projects/classification/sensor/artificial_data/run_experiements.py`
* To plot the data: `nupic.research/projects/classification/sensor/artificial_data/generate_and_plot_data.py`
* To tweak the experiments settings: `nupic.research/projects/classification/sensor/artificial_data/settings.py`

*RES-281*
This is what we talked about yesterday @ywcui1990. To answer your question about how the clusters are merged in my online agglomerative clustering example: the `trim_clusters()` method determines which clusters to keep. It computes the average size of the clusters and then takes 10% of that. For each cluster (at each iteration), if cluster_size > mean_cluster_size * 10% then I keep the cluster. Otherwise, I merge it with the closest one.

I can tweak the 10% values. If I make it 50% it's much more aggressive to merge clusters that are close together, leaving only the biggest ones.

To run the online clustering demo: `nupic.research/projects/classification/sensor/artificial_data/demo_online_clustering.py`


marionleborgne:
Hum. I should have created 2 PRs really... 😓

ywcui1990:
:thumbsup: Is there a reference for this algorithm?

subutai:
@marionleborgne This might also be of interest: 

https://github.com/numenta/nupic.research/blob/master/htmresearch/algorithms/hierarchical_clustering.py

marionleborgne:
@ywcui1990 It's some prior which I documented on my old blog: http://ebrain.io/koac. The papers that are referenced and used as an inspiration for parts of this algo are:
* Agglomerative clustering: "An on-line agglomerative clustering method for non-stationary data". By I. D. Guedalia, M. London, M. Werman.
* Gaussian kernel distance to improve noise robustness: "Improving the robustness of online agglomerative clustering method based on kernel-induce distance measures". By Daoqiang Zhang, Songcan Chen, Keren Tan.


In the end, the logic is simple: it's just online k-means (you can give it the distance function you want) with K set as an upper-bound. At each time step, the method `timclusters()` is called and is taking care of merging clusters that are too small with their closest neighbors - you can set the variable representing what it means for a cluster to be "too small".

marionleborgne:
Thanks @subutai. Some questions:
1. Do you have an example where you used this agglomerative clustering? 
2. Did you use the overlap as the distance metric? If yes, how did it perform with TM states that can be pretty different even if they are part of the same sequence?

Thanks.

ywcui1990:
@marionleborgne Thanks for the refs. 

The hierarchical clustering algorithm @subutai pointed out is developed for the NLP project. The idea is if you train TM on a large set of documents, and then run a hierarchical clustering algorithm on it, maybe we can see some correspondence between the detected clusters and document topics. But I don't think it works well as far as I know. It also doesn't support online clustering.

Here is an example usage in NLP https://github.com/numenta/nupic.research/blob/master/projects/nlp/clustering_runner.py

Here is an unit test:
https://github.com/numenta/nupic.research/blob/master/tests/nlp/unit/hierarchical_clustering_test.py

marionleborgne:
@ywcui1990 thanks for the answers to https://github.com/numenta/nupic.research/pull/587#issuecomment-237939921

subutai:
I didn't implement it, but the reason it didn't work well was because the distance metric and retina we were using didn't work well - it's not an algorithm issue. It doesn't inherently support online clustering although it is probably pretty fast.

marionleborgne:
By the way, this online agglomerative clustering algorithm works in N dimensions. So the example can be extended from 2D vectors to SDRs easily. It's about choosing the right distance metric now... I am working on that.

marionleborgne:
@subutai Yes, I can see how choosing the right distance metric is the meat of the problem. @ywcui1990 and I have been working on visualizing the output of TM against the sequences that we want to classify to design the right distance metric.

PR #588, opened 2016-08-04
[RES-279] Track TM stats over learning in classification network

ywcui1990:
@marionleborgne  Please review

Basic stats of TM will be tracked automatically if
* "temporalImp" is set to "monitored_tm_py" in network config
and
* learning is on for TM

Three stats are being tracked, including
* # predicted -> active cols  (correct predictions)
* # predicted -> inactive cols (extra/false predictions)
* # unpredicted -> active cols (unpredicted inputs)

These stats are updated every 100 iterations

### DIFF
--- a/htmresearch/frameworks/classification/classification_network.py
+++ b/htmresearch/frameworks/classification/classification_network.py
@@ -393,6 +393,19 @@ def trainNetwork(network, networkConfig, networkPartitions, numRecords,
   classifierRegion = network.regions[
     networkConfig["classifierRegionConfig"].get("regionName")]
 
+  trackTMmetrics = False
+  # track TM metrics if monitored_tm_py implementation is being used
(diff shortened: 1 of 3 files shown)


htmresearch/frameworks/classification/classification_network.py
```diff
@@ -459,6 +487,33 @@ def _getClassifierInference(classifierRegion):
 
 
 
+def _inspectTMPredictionQuality(tm, numRecordsToInspect):
+  """Inspect prediction quality of TM over the most recent N records"""
```
marionleborgne: @ywcui1990 By most recent N records, you mean "most recent numRecordsToInspect", right?

ywcui1990: That's right. I have updated the doc string

marionleborgne:
👍 Thanks!

PR #594, opened 2016-08-15
Keeps phases temporal memory working in nupic.research

andrewmalta13:
copies the connections.py that the old temporal_memory_phases.py expects and renames it connections_phases.py.  This is to shield this old implementation from a <a href="https://github.com/numenta/nupic/pull/3254"> recently merged PR </a> that drastically changes connections.py in nupic.


mrcslws:
👍

PR #596, opened 2016-08-16
RES-326: Save network states at each time step along with input data and labels

marionleborgne:


marionleborgne:
@ywcui1990 please review. This is completing RES-326:

```
run_experiment.py should save to CSV, for each timestep:
* record number
* actual value
* label
* TM cells
* UP cells
```

marionleborgne:
The next steps are RES-327 and RES-328 (I started this last one - to add the online clustering)

cc @ywcui1990

htmresearch/frameworks/classification/classification_network.py
```diff
@@ -446,39 +472,68 @@ def trainNetwork(network, networkConfig, networkPartitions, numRecords,
...
+    'tmActiveCellsTrace': tmActiveCellsTrace,
+    'tmPredictiveActiveCellsTrace': tmPredictedActiveCellsTrace,
+    'tpActiveCellsTrace': tpActiveCellsTrace
+  }
+
+  if trackTMmetrics:
+    traces['activeColsTrace'] = activeColsTrace.data
+    traces['predictedActiveColsTrace'] = predictedActiveColsTrace.data
+
+    _plotTMActivation(tm, sensorValueTrace,
```
ywcui1990: I wonder whether it is useful to have this plotting function at all. The plots won't show up on my computer, and the figures are not saved anywhere

ywcui1990:
@marionleborgne I verified that the traces are saved. I think it is good for now. In the long run, I wonder whether we should move the save_traces function out from run_experiments to avoid duplicate code. Assuming that we need to test a different dataset (e.g., sensortag_data), there is no easy way to reuse the save_traces funciton

marionleborgne:
@ywcui1990 I address your 2 comments:
* Removed plotting from the network factory
* Cleaned up `htmresearch/classification/frameworks/utils` and moved the plotting logic into `plot.py` package. You can add the plots there when you get to RES-327 if you want.
* Moved the logic related to saving the traces in a CSV file from `run_experiments.py`  to `htmresearch/classification/frameworks/utils/traces.py`

ywcui1990:
:thumbsup: Please merge

PR #599, opened 2016-08-17
Initial multi-column pooler

subutai:
with basic unit tests


subutai:
@natoromano If you can do a quick scan before I merge that would be great. Thanks!

htmresearch/algorithms/column_pooler.py
```diff
@@ -110,19 +115,17 @@ def compute(self,
...
+    @param learn                    (bool)
+            If True, we are learning a new object
 
     """
     if activeExternalCells is None:
-      activeExternalCells = set()
-
-    if activeApicalCells is None:
-      activeApicalCells = set()
+      activeExternalCells = []
```
natoromano: This should clearly documented somewhere in NuPIC. It caused slight problems in my experiments. Some regions/algorithms are using `set`s, whereas some other use `list`s.

htmresearch/algorithms/column_pooler.py
```diff
@@ -181,15 +180,16 @@ def _computeLearningMode(self, feedforwardInput, lateralInput):
...
+                      activeExternalCells=lateralInput,
+                      formInternalConnections=False,
+                      learn=True)
 
 
   def _computeInferenceMode(self, feedforwardInput, lateralInput):
     """
     Inference mode: if there is some feedforward activity, perform
     spatial pooling on it to recognize previously known objects. If there
-    is no feedforward activity, maintain previous activity.
```
natoromano: You may have deleted a line by inadvertence.

htmresearch/algorithms/column_pooler.py
mrcslws: FWIW the new TM/ETM doesn't use sets, so this awkwardness should go away after we finish updating the Python ETM.

htmresearch/algorithms/column_pooler.py
subutai: Thanks - good catch

htmresearch/algorithms/column_pooler.py
subutai: OK. In this implementation this line should really assign set(). I'll fix it.

htmresearch/algorithms/temporal_memory_phases.py
```diff
@@ -350,7 +350,9 @@ def learnOnSegments(self,
                           self.permanenceDecrement)
 
       if isLearningSegment:
-        n = self.maxNewSynapseCount - len(activeSynapses)
+        n = min(self.maxNewSynapseCount,
```
subutai: @mrcslws Note that I did make the adding synapses bug fix in this implementation, as it is needed to form unions.

mrcslws:
> @mrcslws Note that I did make the adding synapses bug fix in this implementation, as it is needed to form unions.

Ok. This might change the output of the `faulty_temporal_memory`. I don't know if this approach has been tested before, considering the TP had the synapses bug. But it seems right.

PR #615, opened 2016-08-30
Column pooler timing improvements plus cleanup

subutai:


subutai:
@natoromano This is mostly for some small timing improvements and includes your predictedSegmentDecrement change - if you could just check briefly that would be great.

PR #617, opened 2016-08-31
[RES-336] Capacity test of single column L4+L2

ywcui1990:
In this capacity test, we consider a set of objects without any shared (feature, locaiton)
pairs and without any noise. 

In the training phase, a single L4-L2 column is trained on all objects.

In the test phase, we randomly pick a (feature, location) SDR and feed it to
the network. Four metrics are measured
* retrieval accuracy
* the overlap between retrieved L2 SDR and the correct object SDR
* the (max) overlap between retrieved L2 SDR and the incorrect object SDR
* average overlap among L2 object SDRs

In the first experiment, we consider two objects, with varying number of points per object. Using the default parameters of L2-L4 network, a single column allows < 50 points per object

image

In the second experiment, we consider varying number of objects, with fixed number of points per object (10 points). 

image

The result should improve with sampling of L4 cells. Is that implemented?
@subutai @natoromano


ywcui1990:
@subutai I have updated the analysis with subsampling and with a larger network (2048 column of L4). With subsampling, an object can have more (feature, location) pairs, while at the same time it also has slightly more false positive error due to lower threshold. 

However, the increased capacity is less than I expected. For example, from 40 synapses per L4 SDR to 5 synapses per L4 SDR, I didn't see a 8-fold increase of capacity.

subutai:
@ywcui1990 Thanks. Let's discuss these tomorrow.

ywcui1990:
@subutai

I have updated the single column capacity simulation. The results are as expected now. 

In the first experiment, I consider 2 objects with varying number of (feature, location) pairs per object. We can have 100-200 unique locations per object by sampling 5 synapses per L4 input SDR.
image

In the second experiment, I consider varying number of objects, with 10 locations per object. The effect of sampling is more dramatic in this case. We can have ~100 objects with sampling, but much less without sampling. 

image

In both cases, the network starts to have many false positive errors when we go beyond the capacity. This happens because there are too many feedforward connections per L2 cells. A L4 SDR that doesn't belong to the object could falsely activate L2 cells.

PR #618, opened 2016-09-01
RES-344: Complete Python port of ETM

### LINKED ISSUE #598: Update the Python ETM to use the new "walk the columns" strategy
Here's the current status of our temporal memory implementations:

|  | C++ | Python  |
| --- | --- |  --- |
| TemporalMemory | New | New |
| ExtendedTemporalMemory | New | Old (phases) |

The change here is to update the Python ExtendedTemporalMemory to use the new approach. At the end of this, the [ETM algorithm tests](https://github.com/numenta/nupic.research/blob/689d9e3c5575edceda240e83c56a592f4328bd94/tests/extended_temporal_memory/etm_algorithm_test_base.py) should pass, and hopefully the Python and C++ ETM will behave identically.

The new approach walks each column once, doing all cell activation and learning for that column, then moving on to the next. It uses sorted lists of columns, cells, and segments, and it walks them all simultaneously, which avoids doing `set` lookups. It avoids building a pool of "learning segments", making it much easier to understand exactly where synapses are grown / reinforced / punished. This new approach is also easy to parallelize, should we ever want to.

Benefits of this change:

- Faster code
- Cleaner code (according to Marcus)
- Fixes #582 
- Uses the updated Python `Connections` class which supports synapse/segment cleanup.
- Unifies our two ETM implementations so that we can test them by verifying that they get identical results.
- Removes a consumer of `temporal_memory_phases` and `connections_phases`, which we want to remove.

Here's how someone might do this:

1. Copy-paste `temporal_memory.py` into `extended_temporal_memory.py`, and rename it.
2. Change the code the same way that we changed the [C++ TM](https://github.com/numenta/nupic.core/blob/master/src/nupic/algorithms/TemporalMemory.cpp) to the [C++ ETM](https://github.com/numenta/nupic.core/blob/master/src/nupic/experimental/ExtendedTemporalMemory.cpp).
3. Copy-paste `temporal_memory_test.py` into `etm_unit_test.py` and change it to use the ETM instead of TM.

Notes:

- I changed the C++ `Connections` to have a "SegmentExcitationTally" that could be used by the C++ ETM instead of `computeActivity`. I don't think this change will be necessary in the Python -- we can just concatenate lists together via `itertools.chain` ([example](http://stackoverflow.com/a/16346393/396534)) and pass it into `computeActivity`.
- Definitely use our Python `groupby2` function which is already used in `temporal_memory.py`. The current implementation is ready for the ETM.

BoltzmannBrain:
https://jira.numenta.com/browse/RES-344
fixes #598

### DIFF
--- a/htmresearch/algorithms/extended_temporal_memory.py
+++ b/htmresearch/algorithms/extended_temporal_memory.py
@@ -429,14 +429,12 @@ class ExtendedTemporalMemory(object):
     @param random (Object)
     Random number generator. Gets mutated.
     """
-    # TODO: use lambda rather than itertools?
-    # bySegment = lambda segment: segment  #???
-    # for segmentData in groupBy2(cellActiveSegments, bySegment,
-    #                             cellMatchingSegments, bySegment):
-    #   # (segment,
-    #   #  x, y, z) = segmentData  # what is unpacked here?
-    #   segment = segmentData[0]
-    for segment in itertools.chain(cellActiveSegments, cellMatchingSegments):
+    bySegment = lambda segment: segment  #???
+    for segmentData in groupBy2(cellActiveSegments, bySegment,
+                                cellMatchingSegments, bySegment):
+      # (segment,
+      #  x, y, z) = segmentData  # what is unpacked here?
+      segment = segmentData[0]
 
       self.adaptSegment(connections,
                         segment,
@@ -637,7 +635,7 @@ class ExtendedTemporalMemory(object):
   @staticmethod
   def growSynapses(connections, random, segment, numDesiredNewSynapes,
                    internalCandidates, externalCandidates, initialPermanence):
-    """ Creates nDesiredNewSynapes synapses on the segment passed in if
+    """ Creates numDesiredNewSynapes synapses on the segment passed in if
     possible, choosing random cells from the previous winner cells that are
     not already on the segment.
     @param  connections        (Object) Connections instance for the TM
@@ -652,7 +650,8 @@ class ExtendedTemporalMemory(object):
     that was most recently changed is to ensure the same results that we get
     in the c++ implentation using iter_swap with vectors.
     """
-    candidates = itertools.chain(internalCandidates, externalCandidates)
+    candidates = internalCandidates + [connections.numCells() + cell
+                                      for cell in externalCandidates]
 
     # Instead of erasing candidates, swap them to the end, and remember where
     # the "eligible" candidates end
--- a/htmresearch/algorithms/extended_temporal_memory.py
+++ b/htmresearch/algorithms/extended_temporal_memory.py
@@ -159,28 +159,38 @@ class ExtendedTemporalMemory(object):
 
   def compute(self,
               activeColumns,
+              prevActiveExternalCellsBasal,
               activeExternalCellsBasal,
+              prevActiveExternalCellsApical,
               activeExternalCellsApical,
               learn=True):
     """ Feeds input record through TM, performing inference and learning.
 
     @param activeColumns (iter)
     Indices of active columns
+
     @param activeExternalCellsBasal (iter)
     Sorted list of active external cells for activating basal dendrites at the
     end of this time step.
+
     @param activeExternalCellsApical (iter)
     Sorted list of active external cells for activating apical dendrites at the
     end of this time step.
+
     @param learn (bool)
     Whether or not learning is enabled
     """
-    self.activateCells(sorted(activeColumns), learn)
-    self.activateDendrites(learn, activeExternalCellsBasal,
-                           activeExternalCellsApical)
+    self.activateCells(sorted(activeColumns),
+                       prevActiveExternalCellsBasal,
+                       prevActiveExternalCellsApical,
+                       learn)
+    self.activateDendrites(activeExternalCellsBasal,
+                           activeExternalCellsApical,
+                           learn)
 
 
-  def activateCells(self, activeColumns, learn=True):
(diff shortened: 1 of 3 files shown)


BoltzmannBrain:
@mrcslws still a bunch to do (including fixing up all the docstrings) but a skim by you would be appreciated.

I tried to match the c++ ETM logic as much as is reasonable (pythonic); e.g., `adaptSegment` and `growSynapses` take internal and external cells, rather than using `itertools.chain` to combine the internal and external.

Also, I'm unsure how to port [the learning section of `burstColumns`](https://github.com/numenta/nupic.core/blob/master/src/nupic/experimental/ExtendedTemporalMemory.cpp#L636-L647) over; that chunk of `burstColumns` is still todo.

mrcslws:
Here are some notes / explanation on learning in general.

In the C++, the `learnOnCell` function is the common destination for both `activatePredictedColumn` and `burstColumn`. If there are any active segments, it learns on them, or it learns on the best matching segment, or it grows a new segment and learns on it. So `learnOnCell` has 3 different things that it might do.

In the `TemporalMemory`, the learning code is different for the `activatePredictedColumn` and `burstColumn`, but in the `ExtendedTemporalMemory` a big chunk of it is shared (via `learnOnCell`) because:

- A cell can be predicted without having an active apical segment
  - So the `activatePredictedColumn` code has to have a notion of a "best matching apical segment" and it has to grow apical segments -- two things that used to be exclusive to bursting.
- A column can be bursting despite having an active apical segment.
  - So the `burstColumn` code needs to know the "reinforce all active segments" logic.

htmresearch/algorithms/extended_temporal_memory.py
```diff
@@ -21,538 +21,1134 @@
...
+    Random number generator. Gets mutated.
+    """
+    # TODO: use lambda rather than itertools?
+    # bySegment = lambda segment: segment  #???
+    # for segmentData in groupBy2(cellActiveSegments, bySegment,
+    #                             cellMatchingSegments, bySegment):
+    #   # (segment,
+    #   #  x, y, z) = segmentData  # what is unpacked here?
+    #   segment = segmentData[0]
+    for segment in itertools.chain(cellActiveSegments, cellMatchingSegments):
```
mrcslws: I doubt you'll ever want to concatenate that active segments and matching segments.

htmresearch/algorithms/extended_temporal_memory.py
```diff
@@ -21,538 +21,1134 @@
...
+    @param numDesiredNewSynapes (int)    Desired number of synapses to grow
+    @param internalCandidates    (list)   Winner cells in `t-1`
+    @param externalCandidates (list)  Active cells in `t-1`, external to the layer
+    @param  initialPermanence  (float)  Initial permanence of a new synapse.
+
+    Notes: The process of writing the last value into the index in the array
+    that was most recently changed is to ensure the same results that we get
+    in the c++ implentation using iter_swap with vectors.
+    """
+    candidates = itertools.chain(internalCandidates, externalCandidates)
```
mrcslws: For this you'll actually want to create a new list, since the candidates list gets manipulated. The purpose of `itertools.chain` is that it doesn't create a new list, it just creates an iterator that traverses two iterables.

mrcslws: Also, the `externalCandidates` need to have their indices shifted.

Maybe something like:

~~~python
candidates = list(itertools.chain(internalCandidates,
                                  (cell + connections.numCells()
                                   for cell in externalCandidates)))
~~~

BoltzmannBrain:
@mrcslws I tried to get tests to run last night so I can debug what I've done so far, but it looks like the tests need a decent amount of refactoring, for both the changed function signatures (`compute` and on) and to support apical and basal connections. Before I go ahead and do so, I want to confirm that this is correct. The `compute` method now takes four add'l args: `prevActiveExternalCellsBasal`, `activeExternalCellsBasal`, `prevActiveExternalCellsApical`, and `activeExternalCellsApical`. This calls for a lot of changes in the unit tests (b/c `compute` is the TM entry point). So I need to add to every unit test some (list) values for these four args, and also add additional unit tests to cover new scenarios (for basal and apical activities), correct?

mrcslws:
(Quick typing from an airport)

For Python, we could stick with just doing `activeExternalCellsBasal` and `activeExternalCellsApical`, i.e. it doesn't need to take both prev and current. When you use the C++ ETM in Python, it mimics the current Python ETM's method signature: https://github.com/numenta/nupic.core/blob/195cfc8c11f7e2dcb20bd79a46e046f18291b946/src/nupic/bindings/experimental.i#L173

You can also see there how I handled the "formInternalConnections" `compute` param -- I just call the setter on every timestep. (I think we'll want to remove "formInternalConnections" from this signature, it feels silly there. It should be just another TM parameter.)

Also note the [line](https://github.com/numenta/nupic.core/blob/195cfc8c11f7e2dcb20bd79a46e046f18291b946/src/nupic/bindings/experimental.i#L151) `self.connections = self.basalConnections`. So anybody accessing tm.connections automatically gets the basal connections.

I think you'll be able to get the `etm_algorithm_py_test.py` working without changing the test code at all. But `etm_unit_test.py` will be pretty broken. It's just the old "phases" TM unit tests, ported to use the ETM. It doesn't test apical dendrites, etc., and can be replaced with the contents of nupic's `temporal_memory_test.py`, swapping in the ETM.

mrcslws:
These commits have are now part of this PR: #642

PR #628, opened 2016-09-08
Continuous location object machine - first pass

natoromano:
This pull request adds the possibility to experiment with continuous locations on physical objects for the L2-L4 experiments.

This new object machine still sends sensations in the same canonical format to the experiment, but it has a different definition of objects. Objects are now objects from the class `PhysicalObjects`, which should be subclassed with any new object. Those objects have methods to sample from various types of locations (e.g., "face", "edge", and "vertex" for a cube). Each type of location has the same feature everywhere, which is represented, as in the `SimpleObjectMachine`, by an index (the enum of indices are in the base class `PhysicalObject`.

The object machine as well as the physical objects provide methods to represent the objects and sampled locations for sanity checks.

@subutai


natoromano:
@subutai I think it is clearer now.

subutai:
Thanks - this is clearer!

PR #629, opened 2016-09-09
Faster proximal learning in column pooler

mrcslws:
This is for RES-357.

This makes the column pooler about 50% faster, according to `projects/l2_pooling/multi_column_convergence.py`. Running 

~~~python
runExperiment({
  "numObjects": 10,
  "numLocations": 5,
  "numFeatures": 5,
  "numColumns": 7,
  "trialNum": 0
  })
~~~

previously took 19.6 seconds, and now takes 13.1 seconds.

### DIFF
--- a/htmresearch/algorithms/column_pooler.py
+++ b/htmresearch/algorithms/column_pooler.py
@@ -410,7 +410,7 @@ class ColumnPooler(object):
       punish = list(cellNonZeroIndices - activeInputs)
       growthCandidates = activeInputs - cellNonZeroIndices
       newSynapseCount = min(len(growthCandidates), maxNewSynapseCount)
-      grow = sample(growthCandidates, newSynapseCount, self._random)
+      grow = _sample(growthCandidates, newSynapseCount, self._random)
 
       # Make the changes.
       cellPermanencesDense[punish] -= synPermProximalDec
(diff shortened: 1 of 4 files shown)


subutai:
@mrcslws Nice!  Do the tests in tests/algorithms and tests/layers continue to pass?

subutai:
Also tests/regions/column_pooler_region_test.py

htmresearch/algorithms/column_pooler.py
```diff
@@ -532,3 +476,22 @@ def _winnersBasedOnLateralActivity(self,
       predictedActiveCells = predictedActiveCells.union(set(sortedWinnerIndices))
 
     return predictedActiveCells
+
+
+def sample(iterable, k, rng):
```
subutai: Should this be _sample?

mrcslws:
@subutai I replaced the `testPickProximalInputsToLearnOn` since the `_pickProximalInputsToLearnOn` method no longer exists. These new tests cover `_learnProximal` as a whole. They're still unit tests, but they test the units via the `compute` method.

(I've had success with this approach in the TemporalMemory unit tests. Since I added compute-based unit tests to the TM, I've changed the TM many times and the unit tests just keep working. Writing the unit tests this way is more difficult, but the tests survive much longer and they make development more rapid.)

It passes all tests now.

subutai:
👍 Thanks - these look good!

PR #630, opened 2016-09-09
Use predictedActive Cells for clustering analysis

ywcui1990:
@marionleborgne 

Summary
* Ignore empty SDRs when calculating cluster distance
* use Predicted Active Cells for clustering

image

image


marionleborgne:
👍

PR #637, opened 2016-09-15
Generate Neuron Paper figures 6A-6B from current TM implementation

oxtopus:
This reproduces an earlier experiment based on the experimental phases TM implementation in nupic.research, with the production TM implementation in nupic.  Extreme measures have been taken to short-circuit some of the mechanics of TM and Connections to prevent cells that have been destroyed from being used in meaningful ways, allowing for the remaining active cells to pick up the load after cells are killed.

Here are the latest figures this code produces:

figure_6a
figure_6b

Compare to the following image from the paper:

fncir-10-00023-g006

### DIFF
--- a/htmresearch/algorithms/faulty_temporal_memory.py
+++ b/htmresearch/algorithms/faulty_temporal_memory.py
@@ -71,6 +71,12 @@ class FaultyTemporalMemory(TemporalMemory):
 
     print "Total number of segments removed=", numSegmentDeleted
 
+    # Strip out segments for dead cells
+    self.activeSegments = [s for s in self.activeSegments if
+                           s.cell not in self.deadCells]
+    self.matchingSegments = [s for s in self.matchingSegments if
+                             s.cell not in self.deadCells]
+
 
   def burstColumn(self, column, columnMatchingSegments, prevActiveCells,
                   prevWinnerCells, learn):
--- a/htmresearch/algorithms/faulty_temporal_memory.py
+++ b/htmresearch/algorithms/faulty_temporal_memory.py
@@ -23,7 +23,6 @@
 Faulty Temporal Memory implementation in Python.
 """
 import numpy
-from htmresearch.algorithms.temporal_memory_phases import TemporalMemory
 from nupic.research.temporal_memory import  TemporalMemory
 from nupic.research.connections import CellData, Connections
 
--- a/htmresearch/algorithms/faulty_temporal_memory.py
+++ b/htmresearch/algorithms/faulty_temporal_memory.py
@@ -23,25 +23,109 @@
 Faulty Temporal Memory implementation in Python.
 """
 import numpy
-from collections import defaultdict
 from htmresearch.algorithms.temporal_memory_phases import TemporalMemory
+from nupic.research.temporal_memory import  TemporalMemory
+from nupic.research.connections import CellData, Connections
+
+
+
+class FaultyCellData(CellData):
+  """ CellData subclass that adds a `destroyed` property that evaluates to True
+  when a cell is "killed"
+  """
+
+
+  __slots__ = CellData.__slots__ + ["__destroyed"]
+
+
+  def __init__(self):
+    super(FaultyCellData, self).__init__()
+    self.__destroyed = False
+
+
+  @property
+  def destroyed(self):
+    return self.__destroyed
+
+
+  def destroy(self):
+    self.__destroyed = True # Destruction is permanent!
+
+
+
+class FaultyConnections(Connections):
+  def __init__(self,
+               numCells,
+               maxSegmentsPerCell=255,
+               maxSynapsesPerSegment=255):
+    super(FaultyConnections, self).__init__(
+      numCells=numCells,
+      maxSegmentsPerCell=maxSegmentsPerCell,
+      maxSynapsesPerSegment=maxSynapsesPerSegment)
+
+    # Redo _cells assignment to use FaultyCellData, which supports destruction
+    # of cells
+    self._cells = [FaultyCellData() for _ in xrange(numCells)]
+
+
+  @property
+  def cells(self):
+    # Expose otherwise internal _cells as a property
+    return self._cells
+
+
+  def destroyCell(self, cellIdx):
+
+    # Destroy cell
+    self.cells[cellIdx].destroy()
+
+    # Destroy segments.  self.destroySegment() takes care of deleting synapses
+    segmentCount = 0
+    for segment in self.segmentsForCell(cellIdx):
+      self.destroySegment(segment)
+      segmentCount += 1
+
+    return segmentCount
+
+
+  def segmentsForCell(self, cell):
+    # Completely ignore segments for destroyed cells
+    if self.cells[cell].destroyed:
+      return iter([])
+
+    return super(FaultyConnections, self).segmentsForCell(cell)
+
+
+  def synapsesForSegment(self, segment):
+    # Completely ignore synapses on destroyed cells and destroyed segments
+    if self.cells[segment.cell].destroyed or segment._destroyed:
+      return iter([])
+
+    # Return only synapses connected to non-destroyed presynaptic cells
+    return (synapse
+            for synapse
+            in super(FaultyConnections, self).synapsesForSegment(segment)
+            if not self.cells[synapse.presynapticCell].destroyed)
+
+
+  def createSynapse(self, segment, presynapticCell, permanence):
+    # Do NOT create synapses on dead cells/segments or to dead presynaptic cells
+    if (self.cells[segment.cell].destroyed or
+        self.cells[presynapticCell].destroyed or
+        segment._destroyed):
+      return
+
+    return (super(FaultyConnections, self)
+            .createSynapse(segment, presynapticCell, permanence))
+
+
 
 class FaultyTemporalMemory(TemporalMemory):
   """
   Class implementing a fallible Temporal Memory class. This class allows the
   user to kill a certain number of cells. The dead cells cannot become active,
   will no longer participate in predictions, and cannot become winning cells.
-  While admittedly cruel and cold hearted, this feature  enables us to test the
-  robustness of the Temporal Memory algorithm under such situations. Such is the
-  price of progress.
-
-  And by the way, we're not actually killing anything real.
   """
-
-  # ==============================
-  # Main functions
-  # ==============================
-
 
   def __init__(self,
                **kwargs):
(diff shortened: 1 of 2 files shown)


htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -52,12 +135,26 @@ def __init__(self,
...
+    that supports cell destruction.
+
+    See Connections for constructor signature and usage
+
+    @return: FaultyConnections instance
+    """
+    return FaultyConnections(*args, **kwargs)
+
+
   def killCells(self, percent = 0.05):
```
mrcslws: Since this destroys segments on the cell, it should also remove any destroyed segments from `tm.activeSegments` and `tm.matchingSegments`.

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -70,206 +167,139 @@ def killCells(self, percent = 0.05):
...
+      numActive = lambda s: numActivePotentialSynapsesForSegment[s.flatIdx]
 
-          if permanence > 0 and self.predictedSegmentDecrement > 0:
-            numActiveSynapsesForSegment[segment] += 1
+      bestMatchingSegment = None
+      for seg in columnMatchingSegments:
+        if bestMatchingSegment is None:
+          bestMatchingSegment = seg
+        elif seg > bestMatchingSegment and not connections.cells[seg.cell].destroyed:
+          bestMatchingSegment = seg
```
mrcslws: It shouldn't be possible for there to be a matching segment on a dead cell. When we kill a cell, we destroy its segments.

I think burstColumn can be almost identical to the TM's code. Only the "Strip out destroyed cells" part is needed, I think.

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -70,206 +167,139 @@ def killCells(self, percent = 0.05):
...
+  def activatePredictedColumn(self, *args, **kwargs):
+    # Return only non-destroyed cells
+    return [idx
+            for idx
+            in TemporalMemory.activatePredictedColumn(*args, **kwargs)
+            if not self.connections.cells[idx].destroyed]
 
 
-  def computePredictiveCells(self, activeCells, connections):
+  def activateDendrites(self, learn=True):
```
mrcslws: Is it necessary to implement this? As long as there are no segments on the dead cells, the TM's code should work.

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -70,206 +167,139 @@ def killCells(self, percent = 0.05):
...
+    print "Total number of segments removed=", numSegmentDeleted
 
-        if column not in activeColumns:
-          predictedInactiveCells.add(cell)
 
-    return (activeCells,
-            winnerCells,
-            predictedActiveColumns,
-            predictedInactiveCells)
+  def activatePredictedColumn(self, *args, **kwargs):
```
mrcslws: `activatePredictedColumn` won't activate any dead cells, since dead cells can't have active segments. So I don't think this method is needed.

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -23,26 +23,109 @@
...
+  def destroyed(self):
+    return self.__destroyed
+
+
+  def destroy(self):
+    self.__destroyed = True # Destruction is permanent!
+
+
+
+class FaultyConnections(Connections):
```
mrcslws: Is the FaultyConnections needed? It seems it'd be easiest to just keep a set of `self.deadCells` on the `FaultyTemporalMemory`.

Then just never grow a segment from a dead cell, and destroy all segments on a cell when destroying the cell. All this logic can live in the `FaultyTemporalMemory`.

mrcslws:
To use the approach I suggested in my comments, you'll have to copy-paste the activateCells method and have it call the custom burstColumn, also giving it self.deadCells. (This is making up for the fact that burstColumn is not a traditional method that has access to the self)

htmresearch/algorithms/faulty_temporal_memory.py
oxtopus: I didn't think so at first, either.  Without it, I was getting weird results.  For example, it looked great at 75% cell death, but 60% would never recover.  `self.connections.segmentForFlatIdx()` returns segments on destroyed cells, requiring compensation in `activateDendrites()`.

I will follow up by fixing `segmentForFlatIdx()`, which should allow me to remove this function.

htmresearch/algorithms/faulty_temporal_memory.py
oxtopus: `activateDendrites()` is doing that compensation.

htmresearch/algorithms/faulty_temporal_memory.py
oxtopus: I'd added this because at the time, it was.  Some subsequent changes may have fixed that.  I will check again.

htmresearch/algorithms/faulty_temporal_memory.py
mrcslws: activateDendrites calculates the next timestep's segments. I'm talking about the ones that are already calculated.

mrcslws:
Alternately, we could change `burstColumn`, etc. on the TM into traditional methods with short argument lists, and have them immediately call a `_burstColumn` etc. function. Then we wouldn't have to copy-paste `activateCells`. The whole thing could be implemented via `self.burstColumns` and `self.killCells`.

oxtopus:
Thanks for your comments, @mrcslws.  It made me realize there's a bug in `destroyCells()` -- the ordering of operations means that the segments weren't being destroyed!  As such, I can remove a bunch of code that inadvertently compensated for that.  Working on cleaning that up now.

oxtopus:
Ready, again, @mrcslws.  Scope of change has now been narrowed to `killCells()` and `burstColumn()` with the same results.

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -23,32 +23,23 @@
...
   """
   Class implementing a fallible Temporal Memory class. This class allows the
   user to kill a certain number of cells. The dead cells cannot become active,
   will no longer participate in predictions, and cannot become winning cells.
-  While admittedly cruel and cold hearted, this feature  enables us to test the
-  robustness of the Temporal Memory algorithm under such situations. Such is the
-  price of progress.
-
-  And by the way, we're not actually killing anything real.
   """
```
mrcslws: nit: This isn't as funny as it used to be.

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -70,206 +62,57 @@ def killCells(self, percent = 0.05):
...
 
-  def activateCorrectlyPredictiveCells(self,
-                                       prevPredictiveCells,
-                                       prevMatchingCells,
-                                       activeColumns):
+  def burstColumn(self, connections, random, column, columnMatchingSegments,
+                  prevActiveCells, prevWinnerCells,
+                  numActivePotentialSynapsesForSegment, cellsPerColumn,
+                  maxNewSynapseCount, initialPermanence, permanenceIncrement,
+                  permanenceDecrement, learn):
```
mrcslws: Wow! Overriding a class method with a traditional method... I've never thought of that. It only works because the base class calls `self.burstColumn`, not `cls.burstColumn`. This seems a little error-prone -- if the TemporalMemory is ever refactored to call `cls.burstColumn` it will break this (I think).

I think maybe we should change the base TemporalMemory to use regular methods. So the signature will be `burstColumn(self, column, columnMatchingSegments, prevActiveCells, prevWinnerCells)`, and it will immediately call a function `_burstColumn` with the really long argument list. Note this important point: we won't pass in `self.connections`, etc. to this method. The method will access the self and won't contain logic.

We'll will still have long argument lists, so it's not quite the same as https://github.com/numenta/nupic/issues/3309

oxtopus:
Your point is duly noted.  However, long term maintainability is only a
minor concern.  This bit of code is in a repository that is decidedly, and
unapologetically unstable and volatile.  Realistically, I won't be able to
work on this until next week, and there are other pressing issues that keep
getting pushed back.

How important is it, really, that your feedback is addressed before merging?

mrcslws:
I went ahead and made the TM changes. https://github.com/numenta/nupic/pull/3348 . I prefer having my name on the git-blame anyway. 😄 

So we won't have to rely on overriding-a-class-method-with-an-instance-method. I also think it will be much easier to keep this working if it doesn't receive long TM parameter lists on the method, because just adding or removing one parameter will break the subclass. Now these methods get most of their parameters from the `self`, so it will be much more stable.

Our plan is to always keep this script working, since it's linked to in a paper.

oxtopus:
I don't care re: concerns over class methods vs instance methods.  I do recall being part of an earlier proposal, and I agree that it _should_ be instance methods but the proposal failed to achieve consensus.  Regardless, I stand by the approach I've taken here -- it's perfectly reasonable to override a class method with an instance method under the circumstances.  If you disagree, that's ok, just let me know what I need to do to satisfy your preferences so I can get this merged and move on.

htmresearch/algorithms/faulty_temporal_memory.py
oxtopus: Was it _ever_ funny?

mrcslws:
@oxtopus all that remains is:

- Change your `burstColumn` to use the new signature `(self, column, columnMatchingSegments, prevActiveCells, prevWinnerCells, learn)` have it access parameters from the `self`.
- Maybe have `burstColumn` immediately call a `_burstColumn` function which takes its parameters explicitly, to follow the coding style.

In discussion of https://github.com/numenta/nupic/issues/3309, I backed down when I realized that I couldn't solve the problem of "long argument lists", since that was my main focus at the time. But now that I'm focused on this, I see that it's still worth doing instance methods even if we can't solve the long argument lists.

oxtopus:
re: 

> Our plan is to always keep this script working, since it's linked to in a paper.

I think you've just made the case to move it to `nupic`, which is fine by me, but is out of scope for the original task

oxtopus:
@mrcslws This one is ready, again.  Please review.  Thanks!

htmresearch/algorithms/faulty_temporal_memory.py
```diff
@@ -58,6 +49,7 @@ def killCells(self, percent = 0.05):
     time you call this method a permutation list is set up. Calls change the
     number of cells considered dead.
     """
+
```
mrcslws: I think `killCells` needs to edit `self.activeSegments` and `self.matchingSegments`.

Something like

~~~python
self.activeSegments = [s for s in self.activeSegments if s.cell not in self.deadCells]
self.matchingSegments = [s for s in self.matchingSegments if s.cell not in self.deadCells]
~~~

oxtopus: If so, it doesn't affect the performance and results in the same figures as without.

mrcslws: With the code as-is, `activatePredictedColumns` will sometimes activate dead cells. (`burstColumns` will do the right thing.)

PR #640, opened 2016-09-19
[RES-374] Separate threshold parameters in column pooler

ywcui1990:
@subutai 

Implements RES-274

* Introduces separate set of threshold parameters for proximal and distal dendrites
Note that activationThreshold exists only for the distal dendrites, as the proximal activation is determined with spatial pooler like rules.

* Introduces separate set of maxNewSynapsesCount and maxSynapsesCount for proximal and distal dendrites. Typically maxSynapsesCount should be higher on proximal dendrites than the distal dendrite as a single proximal zone has to pool over many L4 inputs

* Make sure number of proximal synapses is less than maxNewProximalSynapseCount. Proximal synapse will stop learning if maxNewProximalSynapseCount is achieved.

* Verify that existing scripts still run after the change.


htmresearch/algorithms/column_pooler.py
```diff
@@ -403,26 +410,89 @@ def _learnProximal(self,
...
-      newSynapseCount = min(len(growthCandidates), maxNewSynapseCount)
-      grow = _sample(growthCandidates, newSynapseCount, self._random)
-
-      # Make the changes.
-      cellPermanencesDense[punish] -= synPermProximalDec
-      cellPermanencesDense[reinforce] += synPermProximalInc
-      cellPermanencesDense[grow] = initialPermanence
-
-      # Update proximalPermanences and proximalConnections.
+      cellNonZeroIndices = list(cellNonZeroIndices)
```
subutai: @ywcui1990 Looks like you are reverting some optimization changes here that were recently made by @mrcslws . Can you first merge with master in your branch and then update the PR?

htmresearch/algorithms/column_pooler.py
```diff
@@ -403,26 +410,89 @@ def _learnProximal(self,
...
+        cellPermanencesDense[newInputs] += initialPermanence
+
+      # Update proximalPermanences and proximalConnections
       proximalPermanences.setRowFromDense(cell, cellPermanencesDense)
       newConnected = numpy.where(cellPermanencesDense >= connectedPermanence)[0]
       proximalConnections.replaceSparseRow(cell, newConnected)
 
 
+
+  def _pickProximalInputsToLearnOn(self, newSynapseCount, activeInputs,
```
subutai: This function was removed in recent changes

htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -544,10 +544,14 @@ def getDefaultL2Params(self, inputSize):
       "synPermProximalInc": 0.1,
       "synPermProximalDec": 0.001,
       "initialProximalPermanence": 0.6,
-      "minThreshold": 10,
+      "minThresholdDistal": 10,
+      "minThresholdProximal": 1,
```
subutai: Why is the default value 1?  That won't be robust to noise, will it?

subutai:
@ywcui1990 I left a few comments. In addition can you make sure tests and the other scripts in projects/l2_pooling continue to run after these changes?   Tests are in tests/algorithms, tests/regions, and tests/layers. You can them from tests using `py.test layers algorithms regions`

ywcui1990:
@subutai  I addressed your comments. I verified that the test passes, except for layers/physical_objects_test.py, which fails before I made the parameter changes for an unrelated reason. I will look at that issue separately.

subutai:
👍

PR #641, opened 2016-09-21
RES-280: Online clustering on accelerometer data

marionleborgne:
_Note: do not merge yet._

I load, shuffle and convert the accelerometer data so that it is in the same format as the artificial data. From the the plot below, it looks like this should be possible to find clusters. My next step is to run the converted accelerometer data though the same network.

acc_x


marionleborgne:
@ywcui1990 The code is ready for review, the network runs on the accelerometer data.
1. generate data with `generate_acc_data.py`
2. run: `run_experiments_v2.py`

The plumbing is there but I am not happy with the results of the clustering, in term of accuracy. I am going to try a couple of different parameters settings (in `clustering_settings.py`). Let me know what you think after looking at this PR and we can discuss. Thanks!

marionleborgne:
My next step is to plot the inter-cluster distance matrix for each activity pair, for example: `walk-run`, `walk-jump`, etc... The idea is to see if we have inter-cluster distances that are big enough to help differentiate between clusters of predictedActiveCells.

htmresearch/frameworks/classification/utils/sensor_data.py
```diff
@@ -33,24 +33,23 @@
...
 
-def plotSensorData(expResults):
+def plotSensorData(inpuFilePaths, markers=True, categoryLabels=None):
   """
   Plot several sensor data CSV recordings and highlights the sequences.
-  @param expResults: (list of dict) list of results for each experiment
   """
 
-  categoryColors = ['grey', 'b', 'r', 'y']
+  categoryColors = ['grey', 'blue', 'yellow', 'red', 'green', 'orange',
```
ywcui1990: If you need a large set of named colors. You could use 
`
from matplotlib import colors
categoryColors  = colors.cnames.keys()`

projects/classification/sensor/artificial_data/clustering_settings.py
```diff
@@ -0,0 +1,22 @@
+# Param for binary data
```
ywcui1990: We should come up a better way to handle multiple sets of parameters

projects/classification/sensor/artificial_data/data/sensortag/jump-5min.csv
```diff
@@ -0,0 +1,1415 @@
+timestamp,x,y,z,g,raw_x,raw_y,raw_z
```
ywcui1990: There is another sensor/sensortag_data/ directory that is parallel with sensor/artificial_data/ We might want to rename the directories to avoid confusion

projects/classification/sensor/artificial_data/generate_acc_data.py
```diff
@@ -0,0 +1,118 @@
...
+  _INFILES = ['walk-5min.csv',
+              'run-5min.csv',
+              'stairs-up-5min.csv',
+              'sit-5min.csv',
+              'jump-5min.csv',
+              'stairs-down-5min.csv',
+              'stumble-5min.csv']
+  _METRICS = ['x', 'y', 'z']
+  _OUTFILE = 'data/sensortag_%s.csv'
+  _SLICES = 2
```
ywcui1990: Each chunk is still pretty large with _SLICE=2. I don't know whether that's good or bad though. It's just an observation.

htmresearch/frameworks/classification/utils/sensor_data.py
marionleborgne: I was using that previously, but it was a bit hard to distinguish adjacent colors, so I ended up using a fixed set.

projects/classification/sensor/artificial_data/generate_acc_data.py
marionleborgne: `SLICES` is set to 10 now. But it's hard to know if 2 or 10 is better since I still need to improve the distance metric to be able to compare classification results. I'll leave it at 10 for now.

projects/classification/sensor/artificial_data/clustering_settings.py
marionleborgne: I agree. I needed to clean up the `classification` section anyway so I ended up refactoring a couple of things: 
* We have 3 sets of settings for 3 separate scripts / phases: `settings/htm_network.py` contains params for running the network with `run_htm_network.py`, regardless of wether we are using accelerometer data or artificial data. `settings/acc_data.py` contains settings for generating the cleaned up accelerometer data files and has clustering params tuned for the accelerometer data. Finally, `settings/artificial_data.py` contains settings for generating the artificial data files and has clustering params tuned for the artificial data.
* Updated the README
* Removed duplicated accelerometer data files

marionleborgne:
@ywcui1990 I did a pretty big cleanup of all folders related to the online sequence classification experiments. I added a parent README as well as other READMEs in some of the sub-folders with more details. 

I will merge this PR, you can review when you are back.

PR #648, opened 2016-10-03
Res 350 capacity of multiple columns

oxtopus:
Renames `single_column_capacity.py` to `capacity_test.py` and allows parameterization of the number of columns and result directories.

The most relevant and important changes specific to supporting multiple columns can be found at https://github.com/oxtopus/nupic.research/blob/7d9ec5dab8d7d0613e702b6ebc6d40e8e46822fa/projects/l2_pooling/capacity_test.py#L181-L195 where test data is split between columns, https://github.com/oxtopus/nupic.research/blob/7d9ec5dab8d7d0613e702b6ebc6d40e8e46822fa/projects/l2_pooling/capacity_test.py#L203-L208 where overlap is computed accounting for multiple columns, everything else is cleanup and infrastructure to support parameterization.


projects/l2_pooling/capacity_test.py
```diff
@@ -165,25 +166,33 @@ def testNetworkWithOneObject(objects, exp, testObject, numTestPoints):
...
   exp._unsetLearningMode()
   exp.sendReset()
 
-  overlap = np.zeros((numTestPoints, numObjects))
+  # Split test points between columns
+  numTestPointsPerColumn = numTestPoints / exp.numColumns
+
+  overlap = np.zeros((numTestPointsPerColumn, numObjects))
+
+  for step, pairs in enumerate(zip(*zip(*[iter(testPairs)] * (len(testPairs) / exp.numColumns)))):
```
subutai: Please add comment explaining the enumerate() expression.

projects/l2_pooling/capacity_test.py
```diff
@@ -165,25 +166,33 @@ def testNetworkWithOneObject(objects, exp, testObject, numTestPoints):
...
+    for colIdx in xrange(exp.numColumns):
```
subutai: Does this mean that a particular column will only see a certain subset of the points? Or can every column see every point?

oxtopus: I've since added a comment above this to explain how it's iterating, but essentially yes.  If there are more than one column, each gets a non-overlapping sequence of feature-location pairs.

subutai: That's a little strange. There's no reason why every column shouldn't get to see every feature-location pair.

PR #655, opened 2016-10-10
Re-add ColumnPooler "compute", add "similar objects" notebook

mrcslws:
Now that I'm actually experimenting with the ColumnPooler in notebooks, I see that it's worth having an easy compute method. This method makes a couple assumptions that are true for all of our current research code. With these assumptions, it's much fewer keypresses to run the ColumnPooler.

It's potentially awkward that the ETM compute method calls activateCells and depolarizeCells in the opposite order that the ColumnPooler calls it. We'll see.

This change also adds a notebook which shows how the ColumnPooler handles objects with many of the same feature-locations.

cc @subutai


subutai:
👍

PR #665, opened 2016-10-28
Some Column Pooler tests weren't testing the proximal connections

mrcslws:
Because we weren't performing a reset after learning, some column pooler tests would pass even if no proximal connections were formed. If they just kept the previous active cells active, they would pass. In some cases, this is what was happening (because the proximal connections were below the connectedPermanence).

/cc @subutai

### DIFF
--- a/tests/algorithms/column_pooler_test.py
+++ b/tests/algorithms/column_pooler_test.py
@@ -101,7 +101,7 @@ class ExtensiveColumnPoolerTest(unittest.TestCase):
     self.init()
 
     object = self.generateObject(1)
-    self.learn(object, numRepetitions=1, newObject=True)
+    self.learn(object, numRepetitions=2, newObject=True)
     # check that the active representation is sparse
     representation = self._getActiveRepresentation()
     self.assertEqual(
@@ -111,6 +111,7 @@ class ExtensiveColumnPoolerTest(unittest.TestCase):
     )
 
     # check that the pattern was correctly learnt
+    self.pooler.reset()
     self.infer(feedforwardPattern=object[0])
     self.assertEqual(
       self._getActiveRepresentation(),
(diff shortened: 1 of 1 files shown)


tests/algorithms/column_pooler_test.py
```diff
@@ -101,7 +101,7 @@ def testLearnSinglePattern(self):
     self.init()
 
     object = self.generateObject(1)
-    self.learn(object, numRepetitions=1, newObject=True)
+    self.learn(object, numRepetitions=2, newObject=True)
```
subutai: Why do we need to increase numRepetitions?

mrcslws: The parameters are:

~~~
      "connectedPermanenceProximal": 0.6,
      "initialDistalPermanence": 0.51,
      "synPermProximalInc": 0.1,
~~~

so each feedforward pattern needs to be seen at least twice before it will reach the `connectedPermanence`. An alternate fix would be to make these tests override these parameters.

mrcslws: Oops, that's the distal permanence... I'm double-checking this now.

subutai: Thanks - makes sense. As a rule I don't think we should override parameters unless we are specifically testing their impact.

mrcslws: The parameters are:

~~~
      "connectedPermanenceProximal": 0.6,
      "initialProximalPermanence": 0.6,
      "synPermProximalInc": 0.1,
      "synPermProximalDec": 0.001,
~~~

The problem was that a single decrement would put it below the connected permanence, so learning two patterns would cause problems. Another fix would be

~~~python
    self.init(overrides={"synPermProximalDec": 0.0})
~~~

subutai: Thanks. I don't want to gratuitously override parameters in tests just to make tests pass. The issue you pointed out could affect a number of simulations and experiments. I'm ok with making numRepetitions=2 or (ideally) improving the parameters so all simulations are more robust. For example, perhaps we should make initialProximalPermanence=0.61 in the column pooler file itself or some other change that could make everything better.

PR #666, opened 2016-10-31
New ColumnPooler algorithm: rely more on distal connections

mrcslws:
This change absorbs the [SetMemory](https://github.com/numenta/nupic.research/blob/abac7c829454eaa4781f6babf4d6779ce687205e/projects/l2_pooling/notebooks/SetMemory_similar_objects.ipynb) algorithm into the ColumnPooler, and it embraces the SparseMatrix.

The proximal and distal synapses now use the same synapse learning code. I made this code significantly faster by adding some features to the SparseMatrix in this PR https://github.com/numenta/nupic.core/pull/1139. Now the learning code looks like this:

~~~python
    permanences.incrementNonZerosOnOuter(
      activeCells, activeInput, permanenceIncrement)
    permanences.incrementNonZerosOnRowsExcludingCols(
      activeCells, activeInput, -permanenceDecrement)
    permanences.clipRowsBelowAndAbove(
      activeCells, 0.0, 1.0)
    permanences.setRandomZerosOnOuter(
      activeCells, activeInput, maxNewSynapseCount, initialPermanence, rng)
~~~

I updated the proximal / distal parameters to be consistent.

This change removes any mention of "predicted cells" from the ColumnPooler, ColumnPoolerRegion, and the monitor mixin. At this point I'm not sure "predicted cells" means anything in the context of Layer 2.

Now that the ColumnPooler isn't using the ExtendedTemporalMemory, I'm backing away from the "depolarizeCells" / "activateCells" interface. The lateral input is always used to calculate lateral support for the current timestep. Callers are responsible for inserting any desired time delay.


mrcslws:
@subutai ready for review.

ywcui1990:
I ran the multi_column_convergence test with the  new algorithm. The results look better than the old algorithm when the column number is high. Nice job @mrcslws 

New algorithm
convergence_1

Old algorithm
convergence_1

mrcslws:
I made the changes suggested in #671. There is no more 'maxNewSynapseCount' in the column pooler. Now there's a 'sampleSizeProximal' and a 'sampleSizeDistal'.

This required another update to nupic.core. I added the changes to the existing PR https://github.com/numenta/nupic.core/pull/1139

mrcslws:
@subutai this is ready for review.

After this PR I'm planning on spending more time improving the "infer_hand_crafted_objects.py" script. Also, that script keeps running into a segfault that is independent of this change. I think we might have C code overrunning a buffer somewhere, corrupting memory. I've been trying to track it down... no luck yet. But I've proved to myself that it's not caused by this PR (or my nupic.core SparseMatrix PR).

mrcslws:
@subutai I made the changes. Ready for review.

htmresearch/algorithms/column_pooler.py
```diff
@@ -404,136 +466,97 @@ def reset(self):
...
+  array = numpy.empty(k, dtype="uint32")
+  rng.sample(numpy.arange(start, end, step, dtype="uint32"), array)
+  return array
+
+
+def _rightVecSumAtNZGtThreshold_sparse(sparseMatrix,
+                                       sparseBinaryArray,
+                                       threshold):
+  """
+  Like rightVecSumAtNZGtThreshold, but it supports sparse binary arrays.
```
subutai: Can you specify here what it actually computes?

htmresearch/algorithms/sensor_placement.py
```diff
@@ -0,0 +1,78 @@
+# ----------------------------------------------------------------------
```
subutai: I don't think this belongs in algorithms. It is pretty specific - should probably go into frameworks/layers

projects/l2_pooling/multi_column_synapse_sampling.py
```diff
@@ -248,11 +245,18 @@ def runExperiment(args):
 
   convergencePoint = averageConvergencePoint(
     exp.getInferenceStats(),"L2 Representation", 40)
+<<<<<<< HEAD
```
subutai: Looks like a bunch of merge conflicts.

htmresearch/algorithms/column_pooler.py
```diff
@@ -83,319 +75,389 @@ def __init__(self,
...
-
-
-  def _activateCellsInferenceMode(self, feedforwardInput):
+      self.activeCells = sorted(_sampleRange(self._random,
+                                             0, self.numberOfCells(),
+                                             step=1, k=self.sdrSize))
+
+    if len(feedforwardInput) > 0:
+      # Proximal learning
+      self._learn(self.proximalPermanences, self._random,
```
subutai: It's really nice that all types of segments can be learned with a single _learn() command.

htmresearch/algorithms/column_pooler.py
```diff
@@ -83,319 +75,389 @@ def __init__(self,
...
-        growthCandidatesExternalBasal=sorted(growthCandidatesExternal),
-        learn=True)
-
-
-  def _activateCellsInferenceMode(self, feedforwardInput):
+      self.activeCells = sorted(_sampleRange(self._random,
+                                             0, self.numberOfCells(),
+                                             step=1, k=self.sdrSize))
+
+    if len(feedforwardInput) > 0:
```
mrcslws: I don't think so? Maybe we want to learn synapses even when there are very few input bits. Usually the minThresholdProximal is applied at the cell level, not the layer level. This is the equivalent of this line in the current implementation. https://github.com/numenta/nupic.research/blob/93ad0cf72a2262a8c1f608bba6b938262774b33b/htmresearch/algorithms/column_pooler.py#L246

htmresearch/algorithms/column_pooler.py
```diff
@@ -40,41 +34,39 @@ class ColumnPooler(object):
...
-
     Parameters:
     ----------------------------
     @param  inputWidth (int)
             The number of proximal inputs into this layer
 
-    @param  lateralInputWidth (int)
-            The number of lateral inputs into this layer
+    @param  lateralInputWidths (list of ints)
+            The number of input bits in each lateral input layer
```
mrcslws: Maybe this is better:

~~~python
"""
    @param  inputWidth (int)
            The number of bits in the feedforward input

    @param  lateralInputWidths (list of ints)
            The number of bits in each lateral input
"""
~~~

I didn't see a reason to impose this limitation. We're free to send in input bits from any source, not necessarily just ColumnPoolers, and not necessarily ColumnPoolers of the same size.

htmresearch/algorithms/column_pooler.py
```diff
@@ -83,319 +75,389 @@ def __init__(self,
...
-                    reinforceCandidatesExternal=(),
-                    growthCandidatesExternal=(),
-                    learn=True):
-    """
+    @param  initialDistalPermanence (float)
+            Initial permanence value for distal synapses
 
-    @param  feedforwardInput (set)
-            Indices of active input bits
+    @param  minThresholdDistal (int)
```
mrcslws: In TemporalMemory terminology, these are both "activation" thresholds. They're not TM "minThresholds" (which should probably be called "learning threshold")

In this implementation, a cell devotes exactly 0 or 1 segments to each L2. When a cell appears in multiple SDRs, or if it learns multiple SDRs in a single lateral input, it will use the same segment for all of them. So having separate "learning" and "activation" thresholds isn't meaningful. This just has the "activation" threshold. I copied the proximal parameter names, but we could choose to change them both to "activationThreshold".

If we want one segment per pattern, we'd have to use the Connections. Using this one-segment approach, I don't think we'll see any scaling issues. We're putting much less stress on the distal segments than we are on the proximal.

PR #672, opened 2016-11-04
Capacity as a function of cortical column number

ywcui1990:
@oxtopus @subutai 

I plotted capacity as a function of cortical column number. There is a small but systemic improvement of capacity when more columns are used. This is because reduced overlap with the false matches, but not because increased overlap with the correct object. The result highlights that lateral connections help disambiguate inputs (reflected as lower overlap with incorrect object"

Note that this result is obtained with the[ new column pooler algorithm](https://github.com/numenta/nupic.research/pull/666). 

image

### DIFF
--- a/projects/l2_pooling/capacity_test.py
+++ b/projects/l2_pooling/capacity_test.py
@@ -34,6 +34,7 @@ import multiprocessing
 import os
 import os.path
 
+import matplotlib as mpl
 import matplotlib.pyplot as plt
 import matplotlib.ticker as ticker
 import numpy as np
@@ -43,7 +44,7 @@ from htmresearch.frameworks.layers.object_machine_factory import (
   createObjectMachine
 )
 from htmresearch.frameworks.layers.l2_l4_inference import L4L2Experiment
-
+mpl.rcParams['pdf.fonttype'] = 42
 
 
 NUM_LOCATIONS = 5000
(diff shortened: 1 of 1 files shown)


ywcui1990:
@oxtopus Could you review my changes to the capacity_test script? Thanks

oxtopus:
This will be helpful in comparing results between differing number of columns and the results shown mirror my observations.  I'm somewhat concerned, however, about the effect of more columns on the result in light of the way accuracy is being calculated and I'm currently experimenting with a couple of options.

For example, here are the results for 1 column vs 8 when using a tie breaker that favors a match as discussed over email in which there is sharper drop off due to the change, but with less of an improvement overall as the number of columns increase: https://app.dominodatalab.com/u/oxtopus/L2-L4-Capacity/compare/581bbb2d3174ed75fe68305c...581ca6473174ed75fe6832ea

Here is an example with a single column and instead of favoring a match, calculates accuracy taking into account the number of matches (the 1/k you suggested): https://app.dominodatalab.com/u/oxtopus/L2-L4-Capacity/view/plots/multiple_column_capacity_varying_object_num_summary.pdf?commitId=00b1120528aa46f55b47ff49eaf41406be89e2c7

I'm working on getting an example with greater number of columns in that scenario for comparison.

PR #675, opened 2016-11-04
Create an example demo of L2 for Cosyne submission

ywcui1990:
@mrcslws @subutai 

I created a simple demo for the Cosyne submission. Can you review this and let me know whether the example makes sense? In the demo I compared a 3-columns network with a single column network side-by-side. Both networks are trained on 10 objects with shared features. After training, I plotted active L2 cells over time during inference. The point is to show that lateral cross-column connections helps disambiguate inputs

Single column. The activity converges after 5 touches:
image

3 column. The activity converges after 1 touch.

image

The number of L2 active cells over time

image

### DIFF
--- a/projects/l2_pooling/cosyne_example.py
+++ b/projects/l2_pooling/cosyne_example.py
@@ -158,13 +158,11 @@ if __name__ == "__main__":
   plt.legend()
   plt.savefig('plots/L2_active_cell_vs_time.pdf')
 
-  numSteps = len(l2ActiveCellsSingleColumn)
-  numL2Cells = exp1.config["L2Params"]['cellCount']
+  fig3 = plotActivity(l2ActiveCellsMultiColumn)
+  st = fig3.suptitle("Three Cortical Column")
+  plt.savefig('plots/L2_active_cell_multi_column.pdf')
 
-  fig1 = plotActivity(l2ActiveCellsMultiColumn)
+  fig1 = plotActivity(l2ActiveCellsSingleColumn)
   st = fig1.suptitle("Single Cortical Column")
   plt.savefig('plots/L2_active_cell_single_column.pdf')
 
-  fig3 = plotActivity(l2ActiveCellsSingleColumn)
-  st = fig1.suptitle("Three Cortical Column")
-  plt.savefig('plots/L2_active_cell_multi_column.pdf')
--- a/projects/l2_pooling/cosyne_example.py
+++ b/projects/l2_pooling/cosyne_example.py
@@ -55,11 +55,12 @@ def plotActivity(l2ActiveCellsMultiColumn):
 
 
 if __name__ == "__main__":
-  numColumns = [1, 3]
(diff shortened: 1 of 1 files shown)


projects/l2_pooling/cosyne_example.py
```diff
@@ -0,0 +1,168 @@
...
+    objectSensations[c] = sensations
+
+  sensationStepsSingleColumn = []
+  sensationStepsMultiColumn = []
+  for step in xrange(len(objectSensations[0])):
+    pairs = [
+      objectSensations[col][step] for col in xrange(numColumns)
+      ]
+    sdrs = objectMachine._getSDRPairs(pairs)
+    sdrSingleColumn = {0: sdrs[0]}
```
mrcslws: nit: `sdrSingleColumn` is unused

projects/l2_pooling/cosyne_example.py
```diff
@@ -0,0 +1,168 @@
...
+  objectId = 0
+  obj = objectMachine[objectId]
+
+  # Create sequence of sensations for this object for all columns
+  numColumns = 3
+  objectSensations = {}
+  for c in range(numColumns):
+    objectCopy = [pair for pair in obj]
+    random.shuffle(objectCopy)
+    # stay multiple steps on each sensation
```
mrcslws: Are we actually staying multiple steps on each sensation? I think we should, but I think this code is just putting each sensation in the list once. Is there a reason not to add these lines? https://github.com/numenta/nupic.research/blob/82ab5114841791f84f91c44ea0e83ed2390937cb/projects/l2_pooling/multi_column_convergence.py#L171-L173

mrcslws:
Nice work.

I think we'll want to somehow illustrate the two time scales: "time steps" and "touches". We should spend multiple time steps on each touch.

In these plots, timestep 0 represents the first touch, timestep 1 the second, etc., so the multi_column experiment is actually taking two touches to converge, but that's likely because we're not spending multiple timesteps on each touch.

Independent of this experiment, I think we might want to choose "which sensor touches which feature-location" less randomly. But this random selection might be fine for this experiment. (With the current approach, multiple sensors will often touch the same feature-location at the same time, and in general the time-to-convergence will be much more random. In my [SetMemory experiment](https://github.com/numenta/nupic.research/blob/master/projects/l2_pooling/notebooks/SetMemory_similar_objects.ipynb) I accounted for this, sometimes choosing to "Move sensors deterministically")

ywcui1990:
@mrcslws Thanks for your feedback. I have included your comments. 

I agree with you that a less random touch scheme will further improve the convergence performance of multiple columns. 

Regarding the distinction between "time steps" and "touches". Are my definitions below correct?

Each touch corresponds to movement to a new (feature, location) on the object.
Each "time step" corresponds to an iteration in the L4-L2 algorithm
The agent can continuously sense the same (feature, location) for multiple time steps.

PR #680, opened 2016-11-19
L2 noise tolerance

mrcslws:
This change adds a script which runs a noise tolerance test on the ColumnPooler.

It also posts a "Fault tolerant L2" notebook that presents a slightly different algorithm. I don't think we should switch to this algorithm, but I think it's useful to have in our back pocket.

In the script, I generate charts varying the following dimensions:

- Choose one of the following approaches to noise:
   - Constant noise: For 20% noise on 40 cells, flip 8 on-bits to off and flip 8 off-bits to on.
   - Varying noise: For 20% mean noise on 40 cells, on each timestep for each column, choose a number around 20% (gaussian with mu 0.2, sigma 0.1). Add this much noise to the feedforward input (in the same way as with the constant noise)
- And then choose which parameter to vary:
  - Column count (Default: 3)
  - Distal sample size (Default: 20)
  - Distal threshold (Default: 13)
  - Proximal sample size (Default: 20)
  - Proximal threshold (Default: 10)

@subutai @ywcui1990 here are the charts. If we want to use one somewhere, we can run more trials on each one if we want to make them smoother. Obviously some of these are silly ideas (like a proximal threshold of 2 -- if I'd tested with more objects or more locations per object that would have lots of false positives)

## Vary column count

image

image

## Vary distal sample size

image

image

## Vary distal threshold

image

image

## Vary proximal sample size

image

image

## Vary proximal threshold

image

image


projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+
+  plotPath = os.path.join("plots",
+                          "varyingNoise_successRate_varyProximalThreshold_%s.pdf"
+                          % time.strftime("%Y%m%d-%H%M%S"))
+  plt.savefig(plotPath, bbox_extra_artists=(lgnd,), bbox_inches="tight")
+  print "Saved file %s" % plotPath
+
+
+if __name__ == "__main__":
+  constantNoise_varyColumns()
```
ywcui1990: Some other scripts demonstrate the same point (e.g., `constantNoise_varyProximalThreshold()` and `varyingNoise_varyProximalSampleSize()`), we can only need one of them here (also see my comments on adding a parameter of noiseSigma)

projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+  poolers = [ColumnPooler(inputWidth=L4_CELL_COUNT,
+                          lateralInputWidths=[4096]*(numColumns-1),
+                          **l2Overrides)
+             for _ in xrange(numColumns)]
+
+  # Learn the objects
+  objectL2Representations = {}
+  for objectName, featureLocations in  objectDescriptions.iteritems():
+    for featureLocationName in featureLocations:
+      for _ in xrange(10):
```
mrcslws: Hmm. 3 is risky, since it will result in distal permanences of exactly 0.6, so a single `synPermDistalDec` will put it below threshold. (since distal learning only occurs on the second and third time step). Yeah, 4 should be enough. I did more because this script isn't really testing the learning rules, so it avoids all risk of accidentally falling below the connected threshold, and focuses on inference. But I'm cool with doing 4 if we want to do the same thing across all our scripts. And, of course, it's faster.

projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+        allLateralInputs = [pooler.getActiveCells() for pooler in poolers]
+        for columnNumber, pooler in enumerate(poolers):
+          feedforwardInput = featureLocationSDRs[columnNumber][featureLocationName]
+          lateralInputs = [lateralInput
+                           for i, lateralInput in enumerate(allLateralInputs)
+                           if i != columnNumber]
+          pooler.compute(feedforwardInput, lateralInputs, learn=True)
```
mrcslws: This is testing L2 in isolation, so the full-blown L2L4 setup isn't needed. And this is superior, since the lateralInputs are properly delayed by 1 timestep.

projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+  for objectName, featureLocations in objectDescriptions.iteritems():
+    for pooler in poolers:
+      pooler.reset()
+
+    sensorPositionsIterator = greedySensorPositions(numColumns, len(featureLocations))
+
+    # Touch each location at least numInitialTouches times, and then touch it
+    # once more, testing it.
+    numTouchesPerTraversal = len(featureLocations) / float(numColumns)
+    numInitialTouches = int(math.ceil(numInitialTraversals * numTouchesPerTraversal))
```
mrcslws: I'm adding some documentation. It traverses the object with the sensors some number of times, then it traverses the object once more. A "traversal" consists of touching every point on the object ~once. Not once per sensor -- that would give an unfair advantage to larger numbers of sensors. Just once, globally. Hence the need to distinguish between a "traversal count" and a "touch count".

projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+          lateralInputs = [lateralInput
+                           for i, lateralInput in enumerate(allLateralInputs)
+                           if i != columnNumber]
+
+          pooler.compute(feedforwardInput, lateralInputs, learn=False)
+
+      if touch >= numInitialTouches:
+        for columnNumber, pooler in enumerate(poolers):
+          activeCells = set(pooler.getActiveCells())
+          inferredCells = objectL2Representations[objectName][columnNumber]
```
mrcslws: Yeah, I'll rename them "correctCells".

projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+  noiseFn = lambda: noiseLevel
+
+  return doExperiment(numColumns, l2Overrides, objectDescriptions, noiseFn,
+                      numInitialTraversals)
+
+
+def doVaryingNoiseExperiment(numColumns, l2Overrides, objectDescriptions,
+                             noiseMu, noiseSigma, numInitialTraversals):
+
+  noiseFn = lambda: random.gauss(noiseMu, noiseSigma)
```
mrcslws: Very good point about setting noiseSigma to 0.

projects/l2_pooling/noise_tolerance.py
```diff
@@ -0,0 +1,809 @@
...
+        r = doConstantNoiseExperiment(numColumns, l2Overrides,
+                                      objectDescriptions, noiseLevel,
+                                      numInitialTraversals=6)
+        results[(numColumns, noiseLevel)].extend(r)
+
+  #
+  # Plot it
+  #
+  numCorrectActiveThreshold = 30
+  numIncorrectActiveThreshold = 10
```
mrcslws: I prefer to keep them local and flexible. There's no correct value for them, different experiments might choose to use different thresholds. I like for the code to imply that.

mrcslws:
@ywcui1990 Good feedback, I changed a lot. Here are some charts. It's interesting seeing what happens as you vary sigma.

# Vary column count and sigma

image

image

image

image

image

# Vary proximal sample size

image

# Vary distal sample size

image

ywcui1990:
Nice results. The new experiment script is much cleaner.

PR #681, opened 2016-11-21
RES-396: Clustering V2

marionleborgne:
* New clustering interface


projects/online_sequence_classification/clustering/clustering_interface.py
```diff
@@ -0,0 +1,275 @@
+import heapq
```
scottpurdy: Need copyright header: [example](https://github.com/numenta/nupic.research/blob/6c2d2f2265ac712e82e609f02eefd020d9096c53/htmresearch/algorithms/TM.py#L2-L9)

projects/online_sequence_classification/clustering/clustering_interface.py
```diff
@@ -0,0 +1,275 @@
...
+import numpy as np
+
+from abc import ABCMeta, abstractmethod
+
+
+
+class Point(object):
```
scottpurdy: This is fine but using a namedtuple might be simpler

projects/online_sequence_classification/clustering/clustering_interface.py
scottpurdy: You might look at Python's builtin PriorityQueue module. It is built on top of `heapq` and is usually an easier interface to use.

marionleborgne:
How to run:
1. **Generate data** with `classification/generate_acc_data.py` or 
`classification/generate_artificial_data.py` 
2. **Run the HTM network** that will generate traces with 
`classification/run_htm_network.py`
3. **Run the clustering** with `clustering/run_clustering.py`

> Note: settings can be found in `classification/settings/` and 
`clustering/settings/`

projects/online_sequence_classification/clustering/clustering_interface.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+  def create_cluster(self, center=None):
+    """
+    Create a cluster.
+    
+    :param center: (Point) point to add.
+    """
+    cluster_id = len(self.clusters) + 1
+    cluster = Cluster(cluster_id, center)
```
ywcui1990: This will create a cluster with size=0 even if center is provided. Is that the desired behavior? If we will always create empty clusters with this function, I suggest to eliminate center from the parameters and rename the function.

projects/online_sequence_classification/clustering/clustering_interface.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    raise NotImplementedError()
+
+
+  def merge_closest_clusters(self):
+    """
+    Merge closest two clusters.
+    """
+    inter_cluster_dists = []
+    for c1 in self.clusters.values():
```
ywcui1990: I noticed that this loop goes through each pair of clusters twice. You could go through each pair once by:
```
numClusters = len(self.clusters)
for i in range(numCluster):
  for j in range(i+1, numClusters):
    c1 = self.clusters.values()[i]
    c2 = self.clusters.values()[j]
```


The same applies to the average_cluster_distance function

projects/online_sequence_classification/clustering/clustering.py
```diff
@@ -1,24 +1,75 @@
...
+# ----------------------------------------------------------------------
+import numpy as np
+from clustering_interface import ClusteringInterface
 
 
-  def add(self, point, actual_label):
-    self.points.append({'point': point, 'label': actual_label})
-    self.size += 1
 
+class PerfectClustering(ClusteringInterface):
```
ywcui1990: Why is this called PerfectClustering?

marionleborgne:
Some additional feedback after discussion with @scottpurdy :
* Clustering interface should be more generic. Expose `learn()` and `infer()`. (`learn` is `add_or_merge` in OnlineClusteringV2 and PerfectClustering)
* Set SDR type to int32 explicitly when loading traces

projects/online_sequence_classification/clustering/clustering_interface.py
marionleborgne: Added missing copyright headers everywhere.

projects/online_sequence_classification/clustering/clustering.py
marionleborgne: Because it gives you perfect cluster predictions.

projects/online_sequence_classification/clustering/clustering.py
```diff
@@ -1,24 +1,75 @@
...
-    self.points = []
-    self.size = 0
+# ----------------------------------------------------------------------
+#  Copyright (C) 2016, Numenta Inc. All rights reserved.
+#
+#  The information and source code contained herein is the
+#  exclusive property of Numenta Inc. No part of this software
+#  may be used, reproduced, stored or distributed in any form,
+#  without explicit written authorization from Numenta Inc.
+# ----------------------------------------------------------------------
```
marionleborgne: This header comes from @scottpurdy 's example: https://github.com/numenta/nupic.research/blob/6c2d2f2265ac712e82e609f02eefd020d9096c53/htmresearch/algorithms/TM.py#L2-L9

@scottpurdy Is that fine or should I replace it with the standard AGPL header? Like https://github.com/numenta/nupic.research/blob/master/projects/capybara/classification/run_htm_network.py#L2-L21

projects/online_sequence_classification/clustering/clustering_interface.py
marionleborgne: Actually, no I shouldn't update that part. I forgot that named tuples are immutable, and I am mutating Point.value (the cluster centroid are points and their value is updated dynamically)

projects/online_sequence_classification/clustering/clustering_interface.py
marionleborgne: Cool. I updated that. It is easier to read now.

projects/online_sequence_classification/clustering/clustering_interface.py
```diff
@@ -0,0 +1,282 @@
...
+    self.points.append(point)
+    self.labels.append(point.label)
+    if self.center is None:
+      self.center = point
+    self.center.value = ((self.center.value * self.size + point.value) /
+                         (self.size + 1))
+    self.size += 1
+
+
+  def merge(self, cluster):
```
marionleborgne: Thanks. Added missing docstring.

projects/online_sequence_classification/clustering/clustering_interface.py
marionleborgne: Good point. Thanks for catching that. I updated the code.

projects/online_sequence_classification/clustering/clustering_interface.py
marionleborgne: It's not the desired behavior, thanks for catching this. I updated the method so that if a center is provided, then it is added to the cluster list of points (the cluster size won't be zero then)

marionleborgne:
@scottpurdy and @ywcui1990 All feedback has been addressed (including @scottpurdy 's feedback in https://github.com/numenta/nupic.research/pull/681#issuecomment-263729855). I'll merge the PR because it's blocking another open PR. Feel free to leave more comments here.

PR #684, opened 2016-11-22
Adds tests for l2_l4_inference

scottpurdy:
fixes RES-356

@lscheinkman please review

The doc strings should be self explanatory. I didn't find any concrete logic to test in isolation (logic is either external to this class or inside the network) so the tests are mainly just validating that the interface works and has some basic sanity checks on the output.


scottpurdy:
ping @lscheinkman

lscheinkman:
👍  LGTM

PR #688, opened 2016-12-07
OPTICS demo

marionleborgne:


marionleborgne:
@ywcui1990 you can try the demo by running `capybara/sandbox/demo_optics.py`. Then open `clustering-structure.html`
newplot

ywcui1990:
Looks great, I am trying your demo now.

ywcui1990:
The PR looks good to me.

marionleborgne:
@ywcui1990 I added a plot of the 2D projection in `capybara/clustering/run_clustering.py`. I was gettign wierd results with tSNE so I adapted the demo `capybara/sandbox/demo_dim_reduction.py` and I compared the result of tSNE and MDS for the 2D projection. I think I have some params problems with tSNE (I think you had a discussion with Mirko about that). I'm adding the projection plots below. what do you recommend for the params settings of tSNE? https://github.com/numenta/nupic.research/pull/688/commits/58ac78603143405623e5cca443d995db96667ace#diff-6ba3b6572c7836e45ee24d1ee7e2536aR121

MDS
mds

tSNE
tsne

ywcui1990:
I observed similar results in my own experiments. Mirko and I discussed about it and the conclusion is tSNE is affective if you have (1) lots of points per class or (2) lots of classes. If you a small number of points per class this happens. I don't know whether this can be fixed by tuning the parameters.

marionleborgne:
@ywcui1990 I see. I tried with 10 classes and 100 SDRs per class and here is what I got: 

tsne-1000_sdrs

It's better. But I still get a better projection with MDS (also 10 classes and 100 SDRs per class). See below. So I am not sure what tSNE brings us? It might be better than MDS when working with dense vectors, but it doesn't seem to do a better job than MDS on sparse vectors. What do you think? 

mds-1000_sdrs

marionleborgne:
After discussion with @ywcui1990 we came to the conclusion that tSNE is only beneficial over MDS if you have (1) lots of sequences and (2) high noise.

marionleborgne:
Here are the results of OPTICS on SDR sequences. The SDRs are projected in 2D with MDS (tSNE didn't work as well here) for visualization purposes.

> Note: Input to OPTICS => Each sequence of SDR is averaged (`np.mean(sdrs, axis=0)`). This is gives us a vector of same dimension as one of the input SDRs. This resulting vector is fed to OPTICS. So if you have 20 sequences of SDRs, you'll have 20 input points to OPTICS.

OPTICS analysis with SDRs:
optics-sdrs

Ground truth labels:
sdrs-ground-truth

marionleborgne:
@ywcui1990 this comment (https://github.com/numenta/nupic.research/pull/688#issuecomment-266839925) summarizes the OPTICS analysis. I will merge this PR for now. We can discuss more when you are back.

PR #690, opened 2016-12-13
RES-395 Implement logging as a decorator for general use

oxtopus:
This moves the logging functionality into a decorator class that can be re-used elsewhere, removes inline logging logic.

### DIFF
--- a/htmresearch/frameworks/layers/l2_l4_inference.py
+++ b/htmresearch/frameworks/layers/l2_l4_inference.py
@@ -79,11 +79,11 @@ projects/layers/multi_column.py
 
 """
 
+import collections
+import cPickle
+import inspect
 import os
 import random
-import collections
-import inspect
-import cPickle
 import matplotlib.pyplot as plt
 from tabulate import tabulate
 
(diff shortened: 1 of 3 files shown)


htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -79,33 +79,105 @@
...
+import random
 import matplotlib.pyplot as plt
 from tabulate import tabulate
 
 from htmresearch.support.register_regions import registerAllResearchRegions
 from htmresearch.frameworks.layers.laminar_network import createNetwork
 
 
+
+class LoggingDecorator(object):
```
subutai: Is it possible to move it to htmresearch/support?  I would also like to use this decorator in the new class I am writing for L2456 networks?

subutai: Please add some comments including a simple example use case.

htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -79,33 +79,105 @@
...
+    """ Returns decorated function that logs calls
+    """
+    def _fn(instance, *args, **kwargs):
+      (_, filename, _, _, _, _) = inspect.getouterframes(inspect.currentframe())[1]
+
+      # Log if, and only if logCalls is set to True, either as an attr on instance
+      # or as a kwarg to the called function (e.g. constructor/__init__) AND
+      # call originated internally
+
+      if (getattr(instance, "logCalls", kwargs.get("logCalls", False)) and
```
subutai: Nice :-)

oxtopus:
@subutai I'm manually testing w/ @lscheinkman's `thing_debug.py`.  I get identical results with one notable caveat: This logs only the exact `args` + `kwarg` that were specified in the call to the functions rather than the result of applying defaults after the function is called.  There's effectively no difference because the defaults are applied at runtime, rather than in the reconstructed call.

I will add documentation and move into support as requested.

oxtopus:
@subutai: one more pass, please!  There are some notable changes to the architecture, but functionality remains basically the same.  By default, it will log everything decorated by `@LoggingDecorator()`, so `sendReset()` has been split into two functions so external calls to `sendReset()` are logged, and internal calls to `_sendReset()` are not.  This seems like a decent compromise after discussing w/ @lscheinkman.

PR #692, opened 2016-12-17
RES-422: Run HTM network on UCI HAR dataset

marionleborgne:

### DIFF
--- a/htmresearch/frameworks/classification/utils/traces.py
+++ b/htmresearch/frameworks/classification/utils/traces.py
@@ -50,15 +50,23 @@ def constructStableIntervals(confidence):
 
 
 def plotTraces(xlim, traces, title, anomalyScoreType,
-               outputFile, numTmCells=None, plotTemporalMemoryStates=False):
+               outputFile, clustering,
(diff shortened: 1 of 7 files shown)


marionleborgne:
With this new longer dataset, we see that the anomaly score decreases over time for each category, which is a good sign. 

Here is an example for category 2 (walking up some stairs) showing the rolling average of the anomaly score over a period of 5 steps:
rolling anomaly score

On the other hand, I am not sure yet what computation to use on the anomaly score to detect the transition between categories. The picture below is an overview of the anomaly score. We can see that when transitioning from category 2 to 1, there is no real increase of the anomaly score. However, within category 2, we see some walking patterns that are a little strange and the anomaly score behaves differently at this time. That's interesting but I am not sure if this is going to lead somewhere... I am going to brainstorm some more about that on Monday. 

several categories

marionleborgne:
The conclusion of this, after a meeting with the team is:
* Use a window of fixed size to take the union of SDRs, instead of using the anomaly score to determine the start/end of this window like proposed originally.
* Rely on the clustering learning mechanism to add or merge new clusters (composed of SDR unions created with the fixed-size window)

PR #694, opened 2016-12-21
ObjectMachine off-by-one bugfix

mrcslws:
`numpy.random` handles its boundaries differently from `random`.


mrcslws:
@ywcui1990 in case you've noticed any weirdness lately with the multi_column_convergence script and 3 features, this is why. It was actually using 2 features.

PR #695, opened 2016-12-21
2 TemporalMemory implementations built with the SparseMatrix

mrcslws:
This PR puts 3 implementations of TemporalMemory into a temporary "sparsematrix_temporal_memory" folder. I'm putting the code here to demo the SegmentSparseMatrix, and so that we can decide where this code should live (and whether it should be in multiple classes).

The SegmentSparseMatrix is being reviewed here: https://github.com/numenta/nupic.core/pull/1185

These implementations are noteworthy because:

- They're written without loops. Every computation is vectorized -- i.e. the loops all live down in C code.
- They don't use the `Connections` class. They store segments in a SparseMatrix.

The three implementations are:

- ~`sequence_memory_with_apical.py`~
  - ~Sequence memory with apical dendrites~
- `basal_context_apical_disambiguation.py`
  - Lateral input + feedback input  with no internal connections
- `basal_context_apical_context.py`
  - Lateral input + feedback input, with no internal connections, and with a new approach to apical dendrites

If you want to skim the code to see the parts dealing with the SegmentSparseMatrix, search for "connections" and "permanences", which are instances of the SegmentSparseMatrix and SparseMatrix, respectively.


mrcslws:
@subutai We can talk about this in-person.

The implementations "sequence_memory_with_apical" and "basal_context_apical_disambiguation" are quite similar. The main reason they are separate is because I didn't want to ever concatenate internal and external inputs. Handling these index offsets is unnecessary complexity that makes the code harder to read and causes bugs. We (currently) never take advantage this -- we only use internal connections exclusively or external connections exclusively. And I think if/when we use both simultaneously, we'll want to handle it explicitly, since the internal activity and external activity will tend to have different statistics, so one will dominate if we're not careful.

So I avoided ever doing offsets by never having internal and external basal connections in a single class.

"sequence_memory_with_apical" is really just "basal_context_apical_disambiguation" with the `self.activeCells` passed in as lateral input. Maybe we should actually just do that. Then we'd just have two classes. And we wouldn't have to decide whether to every concatenate inputs, since the concatenation would happen externally (e.g. in the network API). We could even use the new Network API delayed links to pass the previous active cells in as lateral input.

If we take that approach, it's interesting: the API becomes more functional, and we might just want to have `compute` return `(activeCells, winnerCells, previouslyPredictedCells)` rather than storing them on the `self`.

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
```diff
@@ -0,0 +1,723 @@
...
+    - apicalPotentialExcitations (numpy array)
+      The number of active potential synapses for each apical segment
+      Includes counts for active, matching, and nonmatching segments.
+    """
+
+    basalPermanences = self.basalConnections.matrix
+    apicalPermanences = self.apicalConnections.matrix
+
+    # Active basal
+    basalExcitations = basalPermanences.rightVecSumAtNZGteThresholdSparse(
```
subutai: Consider naming as "basalSegmentOverlaps"
Consider helper function in segmentSparseMatrix for this

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
```diff
@@ -0,0 +1,723 @@
...
+      lateralInput, self.connectedPermanence)
+    activeBasalSegments = np.flatnonzero(
+      basalExcitations >= self.activationThreshold).astype("uint32")
+    self.basalConnections.sortSegmentsByCell(activeBasalSegments)
+
+    # Matching basal
+    basalPotentialExcitations = basalPermanences.rightVecSumAtNZSparse(
```
subutai: "overlap"

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
```diff
@@ -0,0 +1,723 @@
...
+             basalPotentialExcitations,
+             apicalPotentialExcitations):
+    """
+    Adjust synapse permanences, grow new synapses, and grow new segments.
+
+    @param learningActiveBasalSegments (numpy array)
+    @param learningActiveApicalSegments (numpy array)
+    @param learningMatchingBasalSegments (numpy array)
+    @param learningMatchingApicalSegments (numpy array)
+    @param punishedBasalSegments (numpy array)
```
subutai: Not yet punished.

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
```diff
@@ -0,0 +1,723 @@
...
+    """
+    Learn on segments. Reinforce active synapses, punish inactive synapses, and
+    grow new synapses.
+
+    @param permanences (SparseMatrix)
+    @param rng (Random)
+    @param learningSegments (numpy array)
+    @param activeInput (numpy array)
+    """
+    permanences.incrementNonZerosOnOuter(
```
subutai: Maybe comment these two lines clearly

mrcslws:
@subutai I made some of the changes. I'm going to experiment more with exposing the SparseMatrix functionality on a Connections-like class.

Thanh-Binh:
Do you have any c++ version of those implementation?

subutai:
OK, thanks.

mrcslws:
@Thanh-Binh No, I doubt we'll build that any time soon, if ever. The point of this work is to see how fast we can write algorithms in Python while keeping the code readable. Structuring an algorithm like this makes sense for Python, but not for C++. In C++ it doesn't make sense to do the work in batches, since that requires building up transient data structures (e.g. a vector of "learningSegments"). Basically, we already have idealized C++ (Extended)TemporalMemory algorithms. These are idealized Python algorithms.

mrcslws:
@subutai I've done a significant amount of cleanup. There are now two TM classes that have very little duplicated code.

Improvements:

- Created a `SparseMatrixConnections` class which makes the code easier to read and saves us the complexity of having to manually clip permanences.
- Make the `SparseMatrix` use signed integers for synapse counts. Now the Python code doesn't have to worry about negative numbers.
- Add `assumeSorted=False` parameters to methods in the bindings. Now any method that needs sorted inputs will sort the inputs by default, so the Python algorithms don't have to keep numpy arrays sorted (but can choose to).
- Created `htmresearch/support/numpy_helpers.py`. I usually import this as `np2`. I moved some duplicated code of the two TMs into functions in this module.
- Created `htmresearch/algorithms/synapse_learning.py` which holds multiple `SynapseLearningRules` classes and a factory. I like this because:
  - It gives us a good way of sharing the code that takes a set of learning segments / cells and chooses which synapses to grow / reinforce / punish.
  - Since it's not duplicated, we can make it more complicated now. E.g. for some parameter combinations, it's faster to call `connections.growSynapses` rather than `connections.growSynapsesToSample`, but it would be burdensome to do this every time we duplicate the code.
  - It's flexible. I didn't make the Connections class imply that there's a "correct" way to grow synapses.

(Yes, at some point I've got some unit tests to write.)

Regarding subclassing, here's the current status and some tentative thoughts:

- So far I haven't used subclassing. The two TMs inherit from `object`.
- It might be useful to create an abstract base class which shows the common interface.
- I've steered away from sharing functionality by putting it on a base class.
  - I think putting private methods on a base class can imply that there's a "correct" way for a subclass to behave. For example, I wouldn't want a TM base class to imply that subclasses should do synapse sampling.
  - It also isn't ideal for code-sharing. Why only share the code with fellow subclasses of the TM? Why can't a Spatial Pooler reuse this code if it finds it useful?
- Instead I've tried to refactor duplicated code into other modules (e.g. `np2` and `synapse_learning`).

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
```diff
@@ -0,0 +1,482 @@
...
+
+from htmresearch.algorithms.synapse_learning import createSynapseLearningRules
+from htmresearch.support import numpy_helpers as np2
+from nupic.bindings.math import Random, SparseMatrixConnections
+
+
+EMPTY_UINT_ARRAY = np.array((), dtype="uint32")
+
+
+class TemporalMemory(object):
```
subutai: We shouldn't call this class TemporalMemory - name it something more specific.

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
```diff
@@ -0,0 +1,482 @@
...
+  def reset(self):
+    self.activeCells = EMPTY_UINT_ARRAY
+    self.winnerCells = EMPTY_UINT_ARRAY
+    self.prevPredictedCells = EMPTY_UINT_ARRAY
+
+
+  def compute(self,
+              activeColumns,
+              basalInput,
+              basalGrowthCandidates,
```
subutai: For basal and apical, what's the difference between "growth candidates" and the active inputs? Why do we need both?

htmresearch/algorithms/synapse_learning.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @param growthCandidateAxons (numpy array)
+    The (presumably active) input bits that are available for growth
+
+    @param kwargs
+    Additional arguments for specific SynapseLearningRules implementations
+    """
+
+
+class ConnectToActiveAxons(SynapseLearningRules):
```
subutai: Can we just call these "ActiveInputs" instead of axons? The term axon is bit weird in some contexts, such as encoders and spatial pooler.

htmresearch/algorithms/synapse_learning.py
```diff
@@ -0,0 +1,241 @@
...
+    connections.growSynapsesToSample(newSegments, growthCandidateAxons,
+                                     numNewSynapses, self.initialPermanence,
+                                     rng)
+
+
+def createSynapseLearningRules(**kwargs):
+  """
+  Factory for SynapseLearningRules
+  """
+  sampleSize = kwargs.get("sampleSize", -1)
```
subutai: sampleSize and maxSynapsesPerSegment are determining the synapse rule to select. How this is done, and the meaning of those parameters, should be clearly documented.

htmresearch/algorithms/synapse_learning.py
```diff
@@ -0,0 +1,241 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+# See the GNU Affero Public License for more details.
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+
+"""Implementations of various synapse learning rules"""
```
subutai: This file basically controls how you sample from active inputs, and how many to sample. My first impression is that it is a lot of extra complexity (multiple classes, factory method, etc.) and a ton of extra code just for those options. And you still end up with an if-statement in line 209:
```
if self.maxSynapsesPerSegment != -1:
```

I don't see how this file is making the code simpler to understand - as far as I can tell it is taking something conceptually very simple and making it a lot more complex.

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
mrcslws: Growth candidates are the "winner cells". Often we don't want to be able to grow to all active input bits. For example, in sequence memory we only grow to one cell per column. We'll probably want to do the same thing elsewhere, e.g. L2 only connecting to "winner" cells in L4, or L4 only connecting to winners in L6.

htmresearch/algorithms/synapse_learning.py
mrcslws: Fair enough. Yeah, I chose to make it more complicated after factoring it all into this file, since the code wasn't duplicated any more and hence could afford to be more complicated. I was trying to solve the problem of code duplication between the two TMs. (that's why it's in this PR). Maybe I went a little too far.

htmresearch/algorithms/sparsematrix_temporal_memory/basal_context_apical_context.py
subutai: Ah yes - thanks. I forgot about that.

mrcslws:
@subutai Here's a name that I think I like: "ApicalDependentTemporalMemory"

I explain the naming in the docstring:

~~~python
class ApicalDependentTemporalMemory(object):
  """
  An alternate approach to apical dendrites. Every cell SDR is specific to both
  the basal the apical input. Prediction requires both basal and apical support.

  A normal TemporalMemory trained on the sequences "A B C D" and "A B C E" will
  not assign "B" and "C" SDRs specific to their full sequence. These two
  sequences will use the same B' and C' SDRs. When the sequence reaches D/E,
  the SDRs finally diverge.

  With this algorithm, the SDRs diverge immediately, because the SDRs are
  specific to the apical input. But if there's never any apical input, there
  will never be predictions.
  """
~~~

In the other TM I included this blurb in the docstring:

~~~python
class TemporalMemory(object):
  """
  ...
  In other words, the apical connections have no effect unless the basal input
  is a union of SDRs (e.g. from bursting minicolumns).
  ...
  """
~~~

Sorry about the big change. The goal is inherently pretty big: prove the API is good by creating a couple algorithms with it and showing that the code is clean. So I figured anything I could do to make these two files cleaner was in scope for this PR.

PR #699, opened 2017-01-05
Run repeated experiments with SP and test significance of results

ywcui1990:
* Run SP experiment repeatedly with different random dataset
* Plot continuous learning experiments with error bars
* Test the statistical significance of the entropy and noise robustness improvement

Continuous learning with error bars:
image

image


ywcui1990:
@subutai New figures for SP paper revision (to address the concern of statistical significance)

PR #700, opened 2017-01-11
New L2 noise experiment with scatterplots

mrcslws:
@ywcui1990 Here are some tweaks on the existing noise tolerance file. The visualization is generated by [this code](https://github.com/numenta/nupic.research/tree/55e7b32aef768443c04632c359e63c9407b31c42/visualizations/grid-of-scatterplots) which is hosted at http://numenta.github.io/nupic.research/visualizations/grid-of-scatterplots/L2-columns-with-noise.html

I'm gonna try and change that page so that it lets us export the SVG. (But that can be done independent of this PR)

Here's a figure when it's run with 5 trials:

image

### DIFF
--- a/projects/l2_pooling/noise_tolerance_l2_l4.py
+++ b/projects/l2_pooling/noise_tolerance_l2_l4.py
@@ -39,21 +39,42 @@ from htmresearch.frameworks.layers.sensor_placement import greedySensorPositions
 
 
 TIMESTEPS_PER_SENSATION = 3
+NUM_L4_COLUMNS = 2048
 
 
-def withDropout(pattern, dropoutPercent):
+def noisy(pattern, noiseLevel, totalNumCells):
   """
-  Take the input pattern and suppress a portion of its active bits.
+  Generate a noisy copy of a pattern.
 
-  @param pattern (sequence)
-  Sequence of active indices
+  Given number of active bits w = len(pattern),
+  deactivate noiseLevel*w cells, and activate noiseLevel*w other cells.
 
-  @return
-  A new pattern consisting of the previous pattern with some active bits removed
+  @param pattern (set)
+  A set of active indices
+
+  @param noiseLevel (float)
+  The percentage of the bits to shuffle
+
+  @param totalNumCells (int)
+  The number of cells in the SDR, active and inactive
+
+  @return (set)
+  A noisy set of active indices
   """
-  n = int(len(pattern) * (1.0 - dropoutPercent))
+  n = int(noiseLevel * len(pattern))
 
-  return set(random.sample(pattern, n))
+  noised = set(pattern)
+
+  noised.difference_update(random.sample(noised, n))
+
+  for _ in xrange(n):
+    while True:
+      v = random.randint(0, totalNumCells - 1)
+      if v not in pattern and v not in noised:
+        noised.add(v)
+        break
+
+  return noised
 
 
 def createRandomObjects(numObjects, locationsPerObject, featurePoolSize):
@@ -88,7 +109,7 @@ def createRandomObjects(numObjects, locationsPerObject, featurePoolSize):
 
 
 def doExperiment(numColumns, objects, l2Overrides, noiseLevels, numInitialTraversals,
-                 featureDropout, locationDropout):
+                 noisyFeature, noisyLocation):
   """
   Touch every point on an object 'numInitialTraversals' times, then evaluate
   whether it has inferred the object by touching every point once more and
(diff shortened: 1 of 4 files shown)


mrcslws:
@ywcui1990 Alright, I added SVG export to http://numenta.github.io/nupic.research/visualizations/grid-of-scatterplots/L2-columns-with-noise.html

It works in Chrome/Firefox but not Safari.

If you want to try importing one into Illustrator, here's a data file that you can drop onto that page. (Rename it to .json extension if you load it into the visualization page via the file open dialog)
https://www.dropbox.com/s/ss3bfuvf2kmsf6c/varyColumns_sigma0.00_20170111-154627.json?dl=0

I'll be curious to see if Illustrator renders it the same as browsers do. I'm also curious to see if Illustrator becomes sluggish with this chart, since it contains lots and lots of <circle /> elements (one for each dot). Browsers can handle it, so hopefully there's no issue.

ywcui1990:
@mrcslws I exported an svg file from your website and it works with illustrator. Individual components are editable and I can export it to other formats with illustrator. Nice figures!

mrcslws:
I pushed an additional experiment into this PR.

Here are notes on what happens when I:

- Model L4 and do an experiment which suppresses active minicolumns in the L4 feature SDR
- Model L4 and do an experiment which suppresses active bits in the L4 location input

## Drop active minicolumns from L4

Results are similar to the L2-in-isolation test.

image


## Drop bits from the location SDR

**These visualizations are a little bit silly for this task. The script in this PR just creates the data files. We could choose better visualizations if we want to write about this somewhere.**

Whether or not the object is inferred depends very much on the objects themselves. If every object contains a unique bag of features, and no object’s bag is a subset of another object’s bag, then the object will be inferred every time.

Example:

  Object 1: A, B, C
  Object 2: A, B, D
  Object 3: B, C, D

image

It infers every object, with or without location info


But if objects are often different arrangements of the same feature, then the objects won’t tend to be inferred.

Example:

  Object 1: A, B, C
  Object 2: A, C, B
  Object 3: C, A, B

image

After it loses location info, it infers no objects.


Using 10 randomly generated objects with 10 locations each and a feature pool of 10 features, we get > 75% success:

image



Using 10 randomly generated objects with 10 locations each and a feature pool of 3 features, without location info we tend to activate every object SDR:

image



**Multi column**

In an array of sensors, if enough sensors have inferred the object, then all remaining sensors can infer the object without location input.

image


The second column infers the object (because it has location info), so the first column just needs to touch one of the object’s features to infer the object.

projects/l2_pooling/noise_tolerance_l2_l4.py
```diff
@@ -0,0 +1,357 @@
...
+          (column,
+           (locationSDRsByColumn[column][sensorPositions[column]],
+            featureSDRsByColumn[column][features[sensorPositions[column]]]))
+          for column in xrange(1, numColumns))
+
+        for _ in xrange(TIMESTEPS_PER_SENSATION):
+          # Add noise to the first column.
+          featureSDR = featureSDRsByColumn[0][features[sensorPositions[0]]]
+          if featureDropout:
+            featureSDR = withDropout(featureSDR, noiseLevel)
```
ywcui1990: I guess the newly introduced bits won't affect the result much. But it is still good to have them, otherwise it is more like "tuning down signal strength" rather than "add noise".

projects/l2_pooling/noise_tolerance_l2_l4.py
```diff
@@ -0,0 +1,357 @@
...
+
+
+
+if __name__ == "__main__":
+  # Suppress active minicolumns in the L4 feature SDR, observing the impact on
+  # L2. Suppress a constant percent of the minicolumns in one cortical column,
+  # but leave the other cortical columns untouched. Observe what happens as more
+  # untouched columns are added.
+  #
+  # We find that the lateral input from other columns can help correctly active
```
ywcui1990: I find text summaries like this very helpful!

projects/l2_pooling/noise_tolerance_l2_l4.py
```diff
@@ -0,0 +1,357 @@
...
+                 (locationSDRsByColumn[column][location],
+                  featureSDRsByColumn[column][features[location]]))
+                for column in xrange(numColumns))
+           for location in xrange(len(features))])
+         for objectName, features in objects.iteritems()))
+
+  if locationDropout:
+    # Now that the objects are learned, allow bursting L4 minicolumns to
+    # activate cells in L2.
+    exp.L4Columns[0].setParameter("defaultOutputType", 0, "active")
```
mrcslws: This is a good point. At first I agreed with you, but I think we actually we want to isolate this behavior to the noise tolerance tests for now, and leave the other scripts unaffected.

PR #704, opened 2017-01-31
optimize package should be imported seperately

arhik:
It turns out I have to import optimize seperately in this module.
`import scipy` does not import optimize but uses optimize package and error pops when I use scripts which require this module.
`from scipy import optimize` fixed it and I found the reason why at these links:
[docs] (https://docs.scipy.org/doc/scipy/reference/tutorial/general.html#scipy-organization) and [issue] (https://github.com/scipy/scipy/issues/4005) here.


subutai:
👍 Thanks!

PR #705, opened 2017-02-01
Pub 227

subutai:
Updated figures for single column and multi-column convergence charts.


subutai:
@mrcslws Could you take a look and see if you spot any obvious issues?

subutai:
@lscheinkman Threshold fixed.

projects/l2_pooling/multi_column_convergence.py
```diff
@@ -38,54 +40,91 @@
...
+
+    # For each L2 column locate convergence time
+    convergencePoint = 0.0
     for key in stats.iterkeys():
       if prefix in key:
-        itSum += locateConvergencePoint(stats[key], targetValue)
-        itNum += 1
+        convergencePoint = max(
+          convergencePoint,
+          locateConvergencePoint(stats[key], minOverlap, maxOverlap))
```
mrcslws: This approach to finding the convergence point is probably working, but it's a bit fragile. It doesn't actually require that every column has inferred the object at one particular time.

Here's a pathological case: imagine you have 4 columns, and at every point in time only one of the columns has inferred Object A, while the others see a point that is ambiguous between objects A and B.

Touch 1: A|B, A|B, A|B, A
Touch 2: A|B, A|B, A, A|B
Touch 3: A|B, A, A|B, A|B
Touch 4: A, A|B, A|B, A|B

This series of events can happen with our code. And this approach would say that it took 4 touches to converge, when really it never converged.

projects/l2_pooling/multi_column_convergence.py
```diff
@@ -38,54 +40,91 @@
   createObjectMachine
 )
 
-def locateConvergencePoint(stats, targetValue):
+def locateConvergencePoint(stats, minOverlap, maxOverlap):
```
mrcslws: I think these are L2 active cell counts, not overlaps.

projects/l2_pooling/multi_column_convergence.py
```diff
@@ -159,22 +217,40 @@ def runExperiment(args):
 
   # For inference, we will check and plot convergence for each object. For each
   # object, we create a sequence of random sensations for each column.  We will
-  # present each sensation for 3 time steps to let it settle and ensure it
+  # present each sensation for 2 time steps to let it settle and ensure it
```
mrcslws: This number became outdated again. :)

projects/l2_pooling/multi_column_convergence.py
```diff
@@ -275,17 +356,37 @@ def runExperimentPool(numObjects,
...
 
-  Convergence[f,c] = how long it took it to converge with f unique features
-  and c columns.
+  convergence /= numTrials
 
-  Features: the list of features we want to plot
-  """
+  # For each column, print convergence as fct of number of unique features
+  for c in range(2, max(columnRange) + 1):
+    print c, convergence[:, c]
```
mrcslws: nit: now that we run the experiment with 1-column, we can print this for c=1 as well.

projects/l2_pooling/multi_column_convergence.py
subutai: Hmm, this might be a bug. In these cases the L2 columns are supposed to converge. In a clean case like this, once a column has inferred one object unambiguously, it should never increase back up to two.

projects/l2_pooling/multi_column_convergence.py
subutai: Thanks - that's why I introduced a variable, but forgot to update the comment :-)

projects/l2_pooling/multi_column_convergence.py
mrcslws: Independent of whether it's a bug, should we let this script be fooled? I don't think it's super-complicated to find the first touch where every column has ~40 active cells, though the "inferenceStats" approach does seem to complicate things. (rather than just looping over `exp.infer()` and doing the touch counting yourself, and checking exp.getL2Representations() after every touch)

If this is a bug, it's not a bug in the code. Whether a cell is active is determined completely from the feedforward and lateral input (and not from the cell's previous state), except when there's little-to-no feedforward input. If L2 columns could lock in on an object and continue clutching onto it even when it receives more evidence for other objects laterally, then it would be error-prone in situations where you switch the object or when there's noise. You'd get silly situations where 4 columns infer Object B while 1 column stays locked in on Object A. In the current model, the cell's previous state contributes 1 active segment to the lateral support, but it can still be outvoted by other segments.

In the example above, every Object A cell always has 4 active segments while every Object B cell has 3, and 4 segments is not enough to inhibit 3.

subutai:
> If this is a bug, it's not a bug in the code.

I didn't mean to imply it's a code bug - but it's an algorithm issue to think through and potentially an algorithm bug.  

Lateral support does include the column's previous state. Currently we don't deal with switching objects without a reset at all - there are a number of issues that come up with that.  There's also no noise in this situation. Conceptually, we should not get L2 flip flops when we are cleanly sensing a single object.  Perhaps the column's previous state should be given greater weight.

In the current code, I would rather assert if the columns don't settle and flip back and forth - I'll update the code accordingly.

mrcslws:
Yeah, thinking about this more, I can see why we'd consider that an issue. You can construct "optical illusion" objects that can never be inferred by 4+ columns. Currently 3 columns can infer any object, but these issues start becoming possible with 4.

subutai:
Yes, that sounds right. 

I think the best thing for now is to leave it as-is - it's a good alpha implementation. Discussions like these are helping us understand its strengths and weaknesses. We can reconsider these issues in the context of a true continuous learning system that is operating in more realistic scenarios.

PR #706, opened 2017-02-07
Fix active cell initialization bug in column pooler

ywcui1990:
@mrcslws I noticed a bug in the  column pooler. 
During initialization, there is
`self.activeCells = ()` which should correspond to no previously active cells. 

However, if I do this in numpy, 
```
arr = numpy.zeros((10, ))
arr[()] = 1
print arr
```
I filled the entire array with ones. This means at the very first step during inference, the network labeled all cells as "previously active". I tried to fix it in this PR.

### DIFF
--- a/htmresearch/algorithms/column_pooler.py
+++ b/htmresearch/algorithms/column_pooler.py
@@ -543,7 +543,8 @@ def _rightVecSumAtNZGtThreshold_sparse(sparseMatrix,
   an eventual C implementation would.
   """
   denseArray = numpy.zeros(sparseMatrix.nCols(), dtype=GetNTAReal())
-  denseArray[sparseBinaryArray] = 1
+  if len(sparseBinaryArray) > 0:
+    denseArray[numpy.array(sparseBinaryArray)] = 1
   return sparseMatrix.rightVecSumAtNZGtThreshold(denseArray, threshold)
 
 
--- a/htmresearch/algorithms/column_pooler.py
+++ b/htmresearch/algorithms/column_pooler.py
@@ -238,22 +238,19 @@ class ColumnPooler(object):
     prevActiveCells = self.activeCells
 
     # Calculate the feedforward supported cells
-    overlaps = _rightVecSumAtNZGtThreshold_sparse(
-      self.proximalPermanences, sorted(feedforwardInput),
-      self.connectedPermanenceProximal)
+    overlaps = self.proximalPermanences.rightVecSumAtNZGteThresholdSparse(
+      list(feedforwardInput), self.connectedPermanenceProximal)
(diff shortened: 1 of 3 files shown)


htmresearch/algorithms/column_pooler.py
```diff
@@ -543,7 +543,8 @@ def _rightVecSumAtNZGtThreshold_sparse(sparseMatrix,
   an eventual C implementation would.
   """
   denseArray = numpy.zeros(sparseMatrix.nCols(), dtype=GetNTAReal())
-  denseArray[sparseBinaryArray] = 1
+  if len(sparseBinaryArray) > 0:
```
mrcslws: Instead of doing the check, you could just do

~~~python
denseArray[numpy.asarray(sparseBinaryArray)] = 1
~~~

This has the additional benefit of not copying the array when it's already a numpy array.

mrcslws: And it'd be more correct, since any tuple will cause problems, not just an empty tuple.

mrcslws: (Ignore my second comment -- your code solves that problem. But `np.asarray` would be good.)

mrcslws:
Actually, here's a better fix: get rid of this method, and use `permanences.rightVecSumAtNZGteThresholdSparse`. I've since [added this method](https://github.com/numenta/nupic.core/blob/3bbc5c73f48b600ca57352069ba1bf8b06d9ec5d/src/nupic/bindings/sparse_matrix.i#L1704) to the SparseMatrix.

You can change all 3 calls to this method. And there's no need to sort the input.

Also note that I'm saying use "GteThreshold", not "GtThreshold". That's another feature I've recently added.

ywcui1990:
@mrcslws I encountered a problem with both of your suggested fixes that involves `numpy.asarray(sparseBinaryArray)`. It works if `sparseBinaryArray` is a tuple, but it doesn't work if `sparseBinaryArray` is a set. 

In column_pooler_unit_test.py, we allowed `feedforwardInput` and `lateralInputs` to be sets, while `prevActiveCells` is a tuple.

Here is a short code snippet to illustrate the problem
```
import numpy
print numpy.asarray((1,2))
print numpy.asarray(set([1,2]))

```

I can get around this by first converting the `feedforwardInput` to `list` and then do `numpy.asarray`, is this OK?

I can get around

mrcslws:
Hmm. I think this fix is good for now, but I'll make a change after you merge this to make it more performant. (We should avoid making copies of the input. With this code it's going to copy it twice, since we convert it to a list then back to a numpy array.)

I might change it so that the column pooler can't take in sets. That capability is making the column pooler slower and/or more complex.

PR #707, opened 2017-02-07
RES-433 Use delayed links in laminar networks

ywcui1990:
Depends on https://github.com/numenta/nupic.research/pull/706

Summary
* Include propagation delay among L2 columns such that the lateral inputs are synchronized (lateral inputs at time t+1 = active cells at time t)
* Add a test for delayed lateral inputs

@subutai @mrcslws  Please review


htmresearch/frameworks/layers/l2_l4_network_creation.py
```diff
@@ -303,12 +303,21 @@ def createMultipleL4L2Columns(network, networkConfig):
...
+          "L2Column" + suffixSrc, "L2Column" + suffixDest,
+          "UniformLink", "",
+          srcOutput="feedForwardOutput", destInput="lateralInput",
+          propagationDelay=1)
+      elif i > j:
+        network.link(
+          "L2Column" + suffixSrc, "L2Column" + suffixDest,
+          "UniformLink", "",
+          srcOutput="feedForwardOutput", destInput="lateralInput",
+          propagationDelay=0)
```
mrcslws: Is it necessary to do this check? Hopefully we can just always use `propagationDelay=1`.

ywcui1990: Yes, it is necessary. A delay on the link from column i to column j is only appropriate If column i is computed before column j. 

One way to avoid this check is to change the logic in Network.cpp
https://github.com/numenta/nupic.core/blob/master/src/nupic/engine/Network.cpp#L442

We could first prepareInputs for all regions that has the same phase, and then do the compute step. I feel changing that would affect many other things though.

subutai: The idea behind the delay is that it shouldn't matter whether i is computed before j or vice versa. We should be able to have delay=1 for all lateral links.

subutai: Also, there's no guarantee as to which order the columns will be called. The network API is free to call compute on regions in any order as long as they have the same phase.

ywcui1990: Per discussion with @vitaly-krugl  and @scottpurdy, the current implementation is sensitive to the order of the columns. With the current implementation, we need to specify the order of L2 columns and adjust the delay accordingly. 

@vitaly-krugl will change the network class to make it insensitive of orders of regions with the same phase, and we will need to update the logic here after that change.

I have specified the phase of L2 columns to ensure that they will be called in the desired order.

scottpurdy: I created an issue to track this [here](https://jira.numenta.com/browse/NUP-2328). Not necessarily Vitaly that will work it but I will try to get it into the next sprint.

ywcui1990:
@subutai @mrcslws 

This PR is ready for review again. 

Summary:
* With the new delayed links, we no longer need to worry about order of regions
* The lateral connections and apical inputs have a delay=1, no delay for the feedforward inputs
* Test now covers both lateral and apical inputs

subutai:
Please dont merge yet

ywcui1990:
@subutai  Yes, the tests pass with this change. I haven't tested its impact on the convergence performance though.

PR #708, opened 2017-02-07
Fix active cell initialization bug in column pooler

ywcui1990:
* Use `rightVecSumAtNZGteThresholdSparse` method of the sparse matrix class in column_pooler
* Cast `feedforwardInput` and `prevActiveCells` to lists before calculating the overlaps


subutai:
@ywcui1990 Please make sure there is a test that would have caught this bug. Thanks.

ywcui1990:
@mrcslws I find it hard to come up with a test that catches this in the column_pooler. Do you think it makes sense to add some tests in the sparseMatrix class to catch the numpy tuple bug?

mrcslws:
The only way to catch this bug in a test would be to change the `ColumnPooler` to store the `numActiveSegmentsByCell` on the `self`, similar to how in the SP we save the `overlaps` and `boostedOverlaps` to the `self`. Then a test could check that these are correct.

PR #709, opened 2017-02-08
RES-437 Supervised baseline for online sequence classification

marionleborgne:
Baseline model: supervised classification of the output of the temporal memory.

### Setup
* Model: Encoder => SP => TM => Union => Single layer neural net 
* Input: Accelerometer data (x axis)

### How to run
Refer to `projects/capybara/keras/README.md` for instructions of how to run the baseline classifier.

### Results
The classification accuracy on the test set is:
* 71% when classifying directly the TM `predictiveActiveCells` .
* 81% when classifying a union of TM `predictiveActiveCells` (window = 10)
* 92% when classifying a union of TM `predictiveActiveCells` + making the classifier predictions vote over a period of time (window =125)


projects/capybara/baseline/keras/baseline_utils.py
```diff
@@ -0,0 +1,427 @@
...
+import numpy as np
+import pandas as pd
+import os
+
+from sklearn.metrics import accuracy_score
+
+from keras.utils import np_utils
+from keras.models import Sequential
+from keras.layers.core import Dense
+from keras.utils.visualize_util import plot
```
marionleborgne: I have the same version of keras and I don't get this error. It works wether i use theano or tensorflow as a backend.

I think you need to downgrade pydot if you have the latest version, as suggested here: http://stackoverflow.com/questions/38446771/importing-theano-attributeerror-module-object-has-no-attribute-find-graphviz

projects/capybara/classification/settings/htm_network.py
```diff
@@ -26,9 +26,10 @@
 INPUT_DIR = os.path.join(parentDir, 'data')
 OUTPUT_DIR = os.path.join(parentDir, 'results')
 FILE_NAMES = [
-  os.path.join('uci','body_acc_x.csv'),
-  # os.path.join('artificial','binary_ampl=10.0_mean=0.0_noise=0.0.csv'),
-  # os.path.join('artificial','binary_ampl=10.0_mean=0.0_noise=1.0.csv'),
+  os.path.join('uci','body_acc_x_inertial_signals_train.csv'),
+  os.path.join('uci','body_acc_x_inertial_signals_test.csv'),
+  #os.path.join('artificial','binary_ampl=10.0_mean=0.0_noise=0.0.csv'),
+  #os.path.join('artificial','binary_ampl=10.0_mean=0.0_noise=1.0.csv'),
```
marionleborgne: Updated. Note that I am not using the network API anymore. This is part of a performance optimization done after profiling `htm_network.py`.

projects/capybara/datasets/uci_har/README.md
```diff
@@ -48,3 +48,9 @@ https://archive.ics.uci.edu/ml/datasets/Human+Activity+Recognition+Using+Smartph
 
 ## Convert the data
 * Run `convert_to_csv.py`
+
```
marionleborgne: Fixed

projects/capybara/baseline/keras/baseline_utils.py
marionleborgne: @ywcui1990 I removed the keras model visualization. It was a but superfluous since the network is really simple. You should not encounter this error anymore.

marionleborgne:
@ywcui1990 I am done with the changes. I refactored the network so that the dataset is loaded and processed in batches. 

To run:
1. Go to `capybara/htm` and run `htm_runner.py --config=config.yml`. This will process the data in batches and write the HTM traces to disk. The train file is ~900,000 records and test file is ~300,000. At the moment this takes ~23h to run 😭 And this is despite performance improvements listed in [1].
2. Go to `capyabra/baseline` and run `baseline_runner.py`. This will create single layer feedforward network that will be trained on the the HTM traces (the target is the activity, e.g. walking or standing). This takes ~5h to run. Mainly because the input dimension is large (~65K).  I am still looking at how I can make this faster (this is a Keras optimization)
---
[1] Performance improvement changelog:
* If the output the scalar encoder is the same as the previous time step, don't run the HTM. Wait until you get a record that yields a different encoding. This was introduce to prevent the HTM to learn unnecessarily long high-order sequences when the input data is stable, which makes the learning very slow.
* Stop using the network API and use the raw encoder, SP and TM classes. After profiling, @mrcslws pointed rightfully out that there was a lot of conversions from non-zero indices to SDRs (and vice-versa) in the original code that was using the network API. When I save the HTM traces, I am only writing the non-zero indices at each time step. So this is a helpful optimization.
* Updated to the latest nupic that contains a TM optimization [change](https://github.com/numenta/nupic.core/pull/1225) made by @mrcslws .

projects/capybara/htm/htm_network_runner.py
```diff
@@ -0,0 +1,284 @@
...
+    elif self.observedMin > scalarValue:
+      self.observedMin = scalarValue
+
+    if self.observedMax is None:
+      self.observedMax = scalarValue
+    elif self.observedMax < scalarValue:
+      self.observedMax = scalarValue
+
+
+  def _isEncodingChanging(self, encoding):
```
oxtopus: re: RES-437, I think this function does the trick, but it's a little hard to follow having just jumped into this codebase without any prior experience.  Some areas of improvement include:

0. `encoding` is passed as a func arg, but it's compared to an instance property (`self.lastEncoding`) which is set elsewhere.  If the function signature were changed to accept two args: `encoding` and `lastEncoding`, the function is a little easier to interpret.
0. Similarly, instance property `self.encodingRepititionCounter` need not necessarily be an instance property.  Consider limiting scope to `_isEncodingChanging()`. e.g. `self._isEncodingChanging.encodingIsChanging = ...` vs `self.encodingIsChanging`.  Alternatively, you can keep track of encodings in a static var, obviating the need for `encodingRepititionCounter` altogether.

projects/capybara/htm/htm_network.py
```diff
@@ -0,0 +1,245 @@
...
+    spInputWidth = self.encoder.getWidth()
+    self.spParams = {
+      "globalInhibition": True,
+      "columnDimensions": [self.numColumns],
+      "inputDimensions": [spInputWidth],
+      "potentialRadius": spInputWidth,
+      "numActiveColumnsPerInhArea": 40,
+      "seed": 1956,
+      "potentialPct": 0.8,
+      "boostStrength": 0.0,
```
ywcui1990: It may be beneficial to include boosting in most cases. I recommend a small amount of boosting e.g, `boostingStrength=5`

projects/capybara/htm/htm_network.py
```diff
@@ -0,0 +1,245 @@
...
+      "synPermActiveInc": 0.003,
+      "synPermConnected": 0.2,
```
ywcui1990: If you just want to test the value of TM, you may consider turning SP learning off by setting `synPermActiveInc` and `synPermInactiveDec` to zero.

projects/capybara/htm/htm_network.py
```diff
@@ -0,0 +1,245 @@
...
+  def handleRecord(self, scalarValue, label=None, skipEncoding=False,
+                   learningMode=True):
+    """Process one record."""
+
+    if self.runSanity:
+      self.sanity.waitForUserContinue()
+
+    # Encode the input data record if it hasn't already been encoded.
+    if not skipEncoding:
+      self.encoder.encodeIntoArray(scalarValue, self.encoderOutput)
```
ywcui1990: minor: This can be replaced by `self.encodeValue(scalarValue)`

projects/capybara/htm/htm_network.py
marionleborgne: Thanks. Good to know.

htmresearch/frameworks/classification/network_factory.py
```diff
@@ -31,6 +31,13 @@
 from nupic.engine import Network
 from nupic.engine import pyRegions
 
+try:
+  import capnp
+except ImportError:
+  capnp = None
+if capnp:
+  from nupic.proto import NetworkProto_capnp
```
marionleborgne: That's right, good catch. I had initially taken this out of [this example](https://github.com/numenta/nupic/blob/master/tests/integration/nupic/engine/network_creation_common.py#L36-L41) to try out capnp serialization but if `capnp` is unimportable,  `loadNetwork()` and `saveNetwork()` would throw errors. I am going to remove these imports and methods - they are unused because model serialization does not work at the moment.

projects/capybara/baseline/baseline_runner.py
```diff
@@ -0,0 +1,157 @@
...
+PREDICTIONS_FILE = 'predictions.csv'
+MODEL_NAME = 'baseline.h5'
+
+LABELS = ['WALKING',
+          'WALKING_UPSTAIRS',
+          'WALKING_DOWNSTAIRS',
+          'SITTING',
+          'STANDING',
+          'LAYING']
+if __name__ == "__main__":
```
marionleborgne: 👍 Broke it up into functions.

projects/capybara/htm/htm_network.py
marionleborgne: 👍

projects/capybara/htm/htm_network.py
marionleborgne: Thanks. I updated this param.

projects/capybara/htm/htm_network_runner.py
marionleborgne: Thanks for the feedback @oxtopus. I made change # 0.  However, I am not sure what you mean in # 1. I want to keep the flexibility of letting more than 1 repetition happen (hence the `encodingRepetitionCounter`) and I am not sure I can keep this functionality with the change you recommend. I might be missing something though.

marionleborgne:
@oxtopus @ywcui1990 thanks for the feedback, I addressed the comments.

A  note about the next step: After discussing with @subutai and @ywcui1990 it has been agreed to restrict the input dataset to a smaller number of rows (~40,000 instead of ~1,200,000) to speed up model training. The reason for that is that, at the moment, we care more about establishing a supervised baseline for the TM online clustering, than to compare the supervised baseline to other techniques like LSTM. So there is no need to use the full dataset.

I am now running this experiment on the restricted datasets. I will also run the experiment where the TM is disabled and only the output of the SP is used for the supervised baseline.

projects/capybara/htm/htm_network_runner.py
oxtopus: For example, you can use `collections.deque` of max length `encodingRepetitionCounter` with which to buffer recent encodings.   When all encodings in the buffer are consistent, then you can skip processing.

marionleborgne: @oxtopus Thanks for the feedback, again. I made some changes here: https://github.com/numenta/nupic.research/pull/709/commits/418acbc77611bc4eb46dcd5cf14cccbb0f860dec

PR #712, opened 2017-02-18
Update default parameters for L4-L2 network

ywcui1990:
@subutai 

I have updated the default parameter setup for L4-L2 network. 
I also updated the multi_column_convergence experiment to use 150 MCs in L4.

### DIFF
--- a/projects/l2_pooling/multi_column_convergence.py
+++ b/projects/l2_pooling/multi_column_convergence.py
@@ -207,7 +207,7 @@ def runExperiment(args):
   exp = L4L2Experiment(
     name,
     numCorticalColumns=numColumns,
-    inputSize=256,
+    inputSize=150,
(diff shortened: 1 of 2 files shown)


projects/l2_pooling/multi_column_convergence.py
```diff
@@ -207,8 +207,9 @@ def runExperiment(args):
   exp = L4L2Experiment(
     name,
     numCorticalColumns=numColumns,
-    inputSize=2048,
-    numInputBits=40,
+    inputSize=256,
```
subutai: inputSize of 256 is inconsistent with sensorInputSize of 150 above

ywcui1990:
@subutai  Sorry, I made that change locally but forgot to push it to Github. It is 150 now.

PR #713, opened 2017-02-20
[RES-350] Multi-column capacity simulations

ywcui1990:
Summary
* Evaluate capacity of L4-L2 network w.r.t. number of cortical columns.
* Update capacity tests to use small L4 network by default
* Allow change of L2 parameters in each experiment
* Run each experiment multiple times and plot results with error bars

image


ywcui1990:
@subutai Let me know if you have any comments on this PR.

subutai:
👍

PR #714, opened 2017-03-01
Multiple ColumnPooler performance enhancements, ~40% faster

mrcslws:
Changes:

- Clearer/faster `_chooseCells` method. It now makes better use of numpy.
- Use numpy functionality rather than list comprehension to choose which previously active cells to activate.
- Stop allowing sets as feedforwardInput and lateralInput. This allows us to use numpy.asarray on inputs, so that the inputs don't get copied when they're already a numpy array. With this change, all arrays passed into the SparseMatrix are getting used as-is, with no copying / conversion.

Testing the ColumnPooler in isolation, a task that previously took 27.5 seconds now takes 19.1 seconds.

### DIFF
--- a/htmresearch/algorithms/column_pooler.py
+++ b/htmresearch/algorithms/column_pooler.py
@@ -134,7 +134,7 @@ class ColumnPooler(object):
     self.activationThresholdDistal = activationThresholdDistal
     self.distalSegmentInhibitionFactor = distalSegmentInhibitionFactor
 
-    self.activeCells = ()
+    self.activeCells = numpy.empty(0, dtype="uint32")
     self._random = Random(seed)
 
     # These sparse matrices will hold the synapses for each segment.
@@ -150,10 +150,10 @@ class ColumnPooler(object):
     """
     Runs one time step of the column pooler algorithm.
 
-    @param  feedforwardInput (iterable)
+    @param  feedforwardInput (sequence)
             Indices of active feedforward input bits
 
-    @param  lateralInputs (list of iterables)
+    @param  lateralInputs (list of sequences)
             Sets of indices of active lateral input bits, one per lateral layer
 
     @param  learn (bool)
@@ -177,10 +177,10 @@ class ColumnPooler(object):
 
     Parameters:
     ----------------------------
-    @param  feedforwardInput (iterable)
+    @param  feedforwardInput (sequence)
             List of indices of active feedforward input bits
 
-    @param  lateralInputs (list of iterables)
+    @param  lateralInputs (list of sequences)
             Lists of indices of active lateral input bits, one per lateral layer
     """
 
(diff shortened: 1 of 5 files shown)


mrcslws:
I checked a few experiments (multi_column_convergence, noise_tolerance_l2, noise_tolerance_l2_l4) and the results seem to be the same. No changes to the algorithm, as expected.

mrcslws:
@oxtopus would you be up to reviewing this?

The main change is to `_chooseCells`. Also, I'm now requiring inputs to be sorted. This isn't a problem for anyone using the network API since the region uses the ColumnPooler correctly, but it would be possible for someone using the ColumnPooler directly to use it wrong and not know it.

oxtopus:
Will do.

htmresearch/algorithms/column_pooler.py
```diff
@@ -296,34 +302,44 @@ def _chooseCells(self, candidates, n, numActiveSegmentsByCell):
...
+      return numpy.empty(0)
+
+    # Keep a list of numpy arrays.
+    allChosenCells = []
+    numChosen = 0
 
-    chosenCells = []
-    i = 0
+    # Walk the active segment counts in decreasing order.
+    activeSegmentCounts = numpy.unique(
```
oxtopus: Does this statement produce an iterator that iterates in reverse order, or does it create a new array in reverse order?

htmresearch/algorithms/column_pooler.py
```diff
@@ -296,34 +302,44 @@ def _chooseCells(self, candidates, n, numActiveSegmentsByCell):
...
       if activeSegmentCount == 0:
-        chosenCells.extend(orderedCandidates[i:])
+        allChosenCells.append(candidates)
+        numChosen += candidates.size
         break
+      else:
+        # Activate all cells that are not inhibited by this segment count.
+        # A cell is inhibited if another cell has at least
+        # 'distalSegmentInhibitionFactor' times as many active segments.
+        boundary = float(activeSegmentCount) / self.distalSegmentInhibitionFactor
```
oxtopus: Unlikely, but if a user specifies `distalSegmentInhibitionFactor=0` in constructor, is there any risk of a divide-by-zero here?  Would it be beneficial to provide some checks in the constructor?

htmresearch/algorithms/column_pooler.py
mrcslws: It actually creates a "view" on a numpy array. It doesn't copy the array.

`numActiveSegmentsByCell[candidates]` creates an array.
`np.unique` creates an array.
`[::-1]` creates a view.

So yes, you could say that this is effectively an iterator.

htmresearch/algorithms/column_pooler.py
mrcslws: Will do.

PR #718, opened 2017-03-18
Run the same sequence memory tests on different TMs

mrcslws:
This change takes our "tm_algorithm_test_base.py" and makes it more reusable. It then runs these tests on the SparseMatrix TM.

Now any TM implementation can easily hook into these tests. You just have to provide the following methods:

- constructTM
- compute
- reset
- getActiveCells
- getPreviouslyPredictedCells

The test never talks directly to the TM. And it doesn't require a monitor mixin.

The test code now lives in `htmresearch/support/temporal_memory_test`. Eventually I plan on adding a few more files to this folder:

- apical_disambiguation.py
  - I'll refactor existing ETM tests into here.
- apical_context.py
  - None of these tests exist yet. It's for the alternate approach to apical dendrites
- learn_on_one_cell.py
  - I'll refactor existing ETM tests into here.


mrcslws:
(I'm thinking the two TM classes will be called `ApicalTiebreakTemporalMemory` and the `ApicalDependentTemporalMemory`)

I now run the sequence memory tests on both new TMs. For the `ApicalDependentTemporalMemory`, I set the apical input to a constant input, which gives it a constant "world", so the algorithm then just behaves like sequence memory. Then I got cute and switched the inputs around, holding basal input constant and doing sequence memory via apical input. (Not because we plan on doing that -- it's just useful test coverage.)

subutai:
👍 The tests sound good.

PR #719, opened 2017-03-19
NIK merge

subutai:
@lscheinkman I've cleaned up (sort of) and put in all the code including presentation and data files.  Let me know if you have any suggested changes before we merge.


lscheinkman:
It is missing failing on the missing `pkl` file. I will remove from the initial screen so it would work without it if you train the model in Unity first

lscheinkman:
`nik_analysis.py` fails with the following error:
```
$ cat data/train50k_seed37_2.csv | python nik_analysis.py 

Max points exceeded!
Quitting!!!
Number of points= 45001
min/max dx= -0.5216661 0.5216926
min/max dy= -0.52167 0.521676565419
mean dx= 3.34583333333e-05 6.05935618832e-05
mean theta= 44.9383333333 167.019333333
/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/matplotlib/collections.py:548: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if self._edgecolors == 'face'
```

subutai:
> It creates the histogram charts but it displays the message above

Yeah, that's expected. This version only analyzes the first 45,000 points. I changed the message to make it a bit clearer.

ok to merge now?

lscheinkman:
👍 LGTM

PR #720, opened 2017-03-19
3d format .ply as physical object

arhik:
3D formatted files with extension `.ply` can be easily loaded now. I hope this helps you in further research.

- depends on `plyfile` package installable using `pip install plyfile`. 
- `.ply` is easy format to choose and write our own or create/edit using blender/meshlab etc.
- Complex physical objects can be imported for analysis.
- Visualizer is included for realtime update programming. Not necessary but this visualizer aided me in the correct direction.
- Matplotlib surface plots are clean too.
- [example ply files repo to use](http://people.sc.fsu.edu/~jburkardt/data/ply/ply.html)

Note:

- PhysicalObject.SURFACE is added... i forgot to remove it. But it still makes sense to keep it. Can stand for any curved surface.

TODO:

- `SURFACE` case is not programmed yet. This might need computational geometry algorithms, kdtree etc..
- The idea is to flag any face with adjacent area normals `np.dot(n1,n2) >= 1.0 - normalTolerance` with `normal tolerance = .1` per say can be used and flag it as `SURFACE` instead of face and ignore edges because its smooth.


arhik:
####  Result when sampled only from vertices
octahedron_edge

#### Result when sampled from faces
octahedron_face

Results on Matplotlib

image

image

subutai:
@arhik Thank you. We are not using .ply files now, but perhaps this will be useful later. Are you using this code at all? Just curious. Also, can you merge in latest from master before we merge this?

arhik:
@subutai. sure.

I may not use this code. I am just trying to contribute.

HTM is an obsession and I usually try to understand code regarding layers whenever I am free.  I might try to run the code with different models and then compare from higher level point of view. I will share results if I ever do it.

Studying translation invariance and rotational invariance properties were in my head while going through code. 

Just a quick question. In case you will prefer .ply files how will you want the learningConfig be formatted (In continuous object machine code per say)?

arhik:
@subutai Ignore my question. I was confused for some reason. learningConfig doesnot need any changes. 'Continuous location object machine' is compatible with PlyModel class so far.

subutai:
👍

PR #723, opened 2017-03-24
Adding 3 new kinds of noise

ThomasMiconi:


htmresearch/frameworks/layers/feedback_experiment.py
```diff
@@ -255,6 +254,10 @@ def infer(self, sequence, reset=True, sequenceNumber=None, burnIn=2,
       self.sensorInputs[0].addDataToQueue(list(s), 0, 0)
       self.network.run(1)
 
+      # DEBUGGING
```
subutai: Please take out commented code.

PR #724, opened 2017-03-29
RES-492: Add object classification method

lscheinkman:
@subutai Please review


htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -578,6 +580,42 @@ def getL2Representations(self):
...
+      score = 0.0
+      for i in xrange(self.numColumns):
+        # Ignore inactive column
+        if len(l2sdr[i]) == 0:
+          continue
+
+        count += 1
+        overlap = len(l2sdr[i] & objectSdr[i])
+        if overlap >= minOverlap:
+          score += overlap
```
subutai: I think this should just be += 1. The number of bits doesn't matter once it's above a threshold.

htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -578,6 +580,42 @@ def getL2Representations(self):
...
+
+      if count == 0:
+        results[objectName] = 0
+      else:
+        results[objectName] = score / (count * sdrSize)
```
subutai: Just divide by `count`

htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -578,6 +580,42 @@ def getL2Representations(self):
     return [set(column._pooler.getActiveCells()) for column in self.L2Columns]
 
 
+  def getCurrentClassification(self, minOverlap=None):
+    """
+    A dict with a score for each object. Score goes from 0 to 1. A 1 means
```
subutai: Should we return a numpy array, or would that make it hard for Unity?

lscheinkman: A dictionary is easier to convert to JSON for Unity

PR #725, opened 2017-03-29
Shared "apical tiebreak" tests

mrcslws:
Two changes:

### Port apical tests into two reusable suites

One suite focuses on using apical input for sequence memory.

  - I reworked the sequence tests so that the asserts are very precise and so that shared subsequences don't mess up the tests.
  - I didn't port the temporal noise tests because they're redundant with these tests. The "BCDE" tests are essentially a superset of all temporal noise tests.
    - The temporal noise tests are still running on the ETM. (Except for the ones that are already disabled)

The other suite focuses on passing in external basal / apical input.

  - This suite is totally new, and better reflects our current usage of apical dendrites.

  
### More descriptive filenames. Stop relying on folder names.

Life is easier as a developer when you can quickly identify a file by its filename. With my previous scheme, the file paths are informative, but if you only see the filename, it's difficult to work with.

In the new scheme:

- An example test base: `htmresearch/support/shared_tests/apical_tiebreak_test_base.py`
- An example test: `tests/extended_temporal_memory/etm_apical_tiebreak_test.py`
  - Currently it's `etm_apical_tiebreak_cpp_test.py` -- eventually we'll get rid of the Python ETM and then rename these files.


tests/apical_tiebreak_temporal_memory/attm_apical_tiebreak_sequences_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+import numpy as np
+
+from htmresearch.algorithms.sparsematrix_temporal_memory.basal_context_apical_disambiguation import (
+  ApicalTiebreakTemporalMemory)
+from htmresearch.support.shared_tests.apical_tiebreak_sequences_test_base import (
+  ApicalTiebreakSequencesTestBase)
+
+
+class ApicalTiebreakTM_ApicalTiebreakSequenceMemoryTests(ApicalTiebreakSequencesTestBase,
```
subutai: Hard to understand these long names - a descriptive comment would be helpful.

tests/apical_tiebreak_temporal_memory/attm_apical_tiebreak_test.py
```diff
@@ -0,0 +1,83 @@
...
+
+import numpy as np
+
+from htmresearch.algorithms.sparsematrix_temporal_memory.basal_context_apical_disambiguation import (
+  ApicalTiebreakTemporalMemory)
+from htmresearch.support.shared_tests.apical_tiebreak_test_base import (
+  ApicalTiebreakTestBase)
+
+
+class ApicalTiebreakTM_ApicalTiebreakTests(ApicalTiebreakTestBase,
```
subutai: Hard to understand these long names - a descriptive comment would be helpful.

htmresearch/support/shared_tests/apical_tiebreak_sequences_test_base.py
```diff
@@ -0,0 +1,418 @@
...
+           "======================================================\n"
+    ).format(self.id(), self.shortDescription())
+
+
+  def testSequenceMemory_BasalInputRequiredForPredictions(self):
+    """
+    Learn ABCDE with F1.
+    Reset, then observe B with F1.
+
+    It should burst, despite the fact that the B cells have apical support.
```
subutai: Nice comments for each test case!

htmresearch/support/shared_tests/apical_tiebreak_sequences_test_base.py
```diff
@@ -0,0 +1,418 @@
...
+
+    self.assertEqual(eCells, set(self.getActiveCells()))
+    self.assertEqual(eCells, set(self.getPreviouslyPredictedCells()))
+
+
+  def testSequenceMemory_IncorrectFeedbackLeadsToBursting(self):
+
+    """
+    Train on ABCDE with F1, XBCDY with F2.
+    Test with BCDE with F2. Y should burst.
```
subutai: You mean E should burst?

mrcslws:
@subutai I made the changes. This PR also has some "previouslyPredictedCells" that I'll rename after this is merged into master and then into my other PR.

PR #726, opened 2017-03-29
Fix the regions to match API change in nupic.core

mrcslws:
Currently we get `TypeError: initialize() takes exactly 3 arguments (1 given)` when running any L2-L4 experiments.

This change matches https://github.com/numenta/nupic/pull/3494, but for nupic.research

FYI @oxtopus


oxtopus:
👍

PR #727, opened 2017-03-30
Use the same L4->L2 connection for both inference/learning

mrcslws:
Stop toggling the "defaultOutputType" on L4. Instead, distinguish between "feedforwardInput" and "feedforwardGrowthCandidates" in L2.

This requires a small change in the learning logic. `increaseRowNonZeroCountsOnOuterTo` is no longer useful because the number of new synapses should be calculated from the number of existing synapses to active inputs, not from the number of existing synapses to "growth candidate" inputs.

This change depends on https://github.com/numenta/nupic.core/pull/1263

### DIFF
--- a/htmresearch/algorithms/column_pooler.py
+++ b/htmresearch/algorithms/column_pooler.py
@@ -148,7 +148,8 @@ class ColumnPooler(object):
                                    for n in lateralInputWidths)
 
 
-  def compute(self, feedforwardInput=(), lateralInputs=(), learn=True):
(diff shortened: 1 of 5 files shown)


subutai:
Nice idea. We might as well get rid of defaultOutputType in ExtendedTMRegion as well ??  Also the outputs feedForwardOutput and activeCells are redundant now.

mrcslws:
@subutai Agreed. I made those exact changes in the other PR. :)  https://github.com/numenta/nupic.research/pull/729

PR #729, opened 2017-03-30
Add an ApicalTMRegion and integrate it into L2L4 experiments

mrcslws:
This change is dependent on https://github.com/numenta/nupic.research/pull/727

Now you can use the "loopless" TM by creating the L4L2Experiment with L4RegionType="py.ApicalTMRegion". It works with the multi_column_convergence experiment and others.

This also changes the ExtendedTMRegion so that the two have the same input/output names so that it's easy to swap between them in the L2L4 experiments.


htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -236,7 +242,7 @@ def __init__(self,
 
     self.sensorInputs = []
     self.externalInputs = []
-    self.L4Columns = []
+    self.L4Regions = []
     self.L2Columns = []
```
subutai: Should be consistent naming - L2Regions

htmresearch/frameworks/layers/l2_l4_inference.py
```diff
@@ -561,14 +567,16 @@ def getL4Representations(self):
...
+    return [set(column.getOutputData("activeCells").nonzero()[0])
+            for column in self.L4Regions]
 
 
   def getL4PredictiveCells(self):
     """
     Returns the predictive cells in L4.
     """
-    return [set(column._tm.getPredictiveCells()) for column in self.L4Columns]
+    return [set(column.getOutputData("previouslyPredictedCells").nonzero()[0])
```
subutai: Don't we want cells that are currently in predicted state?

htmresearch/regions/ApicalTMRegion.py
```diff
@@ -0,0 +1,437 @@
...
+                          "whole apicalInput is used."),
+          "dataType": "Real32",
+          "count": 0,
+          "required": False,
+          "regionLevel": True,
+          "isDefaultInput": False,
+          "requireSplitterMap": False},
+      },
+      "outputs": {
+        "previouslyPredictedCells": {
```
subutai: Confusing to call this previouslyPredictedCells

PR #731, opened 2017-03-30
Updated tests

subutai:
Fixed testSimpleExperiment
Added testConsistency

### DIFF
--- a/tests/layers/l2_l4_inference_test.py
+++ b/tests/layers/l2_l4_inference_test.py
@@ -22,9 +22,19 @@
 """Tests for l2_l4_inference module."""
 
 import unittest
+import numpy
 
 from htmresearch.frameworks.layers import l2_l4_inference
 
+
+def _randomSDR(numOfBits, size):
+  """
+  Creates a random SDR for the given cell count and SDR size
+  :param numOfBits: Total number of bits
+  :param size: Number of active bits desired in the SDR
+  :return: list with active bits indexes in SDR
+  """
+  return list(numpy.random.permutation(numOfBits)[0:size])
 
 
 class L4L2ExperimentTest(unittest.TestCase):
@@ -34,7 +44,7 @@ class L4L2ExperimentTest(unittest.TestCase):
   and the real work is all done inside the network. The tests here make sure
   that the interface works and has some basic sanity checks for the experiment
   statistics. These are intended to make sure that the code works but do not go
-  far enought to validate that the experiments are set up correctly and getting
+  far enough to validate that the experiments are set up correctly and getting
   meaningful experimental results.
   """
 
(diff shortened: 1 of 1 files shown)


subutai:
@mrcslws Could you take a look at this failure in testConsistency?
CC: @lscheinkman

tests/layers/l2_l4_inference_test.py
```diff
@@ -22,10 +22,20 @@
...
 
 
+def _randomSDR(numOfBits, size):
+  """
+  Creates a random SDR for the given cell count and SDR size
+  :param numOfBits: Total number of bits
+  :param size: Number of active bits desired in the SDR
+  :return: list with active bits indexes in SDR
+  """
+  return list(numpy.random.permutation(numOfBits)[0:size])
```
mrcslws: Have you considered `random.sample(xrange(numOfBits), size)`? I expect it to be dramatically more efficient. (This doesn't address the test failure)

subutai: yes good point - always forget about that

tests/layers/l2_l4_inference_test.py
```diff
@@ -192,6 +204,139 @@ def testCapacity(self):
...
+      sensations = [
+          objectsToLearn["Mug"][0]
+      ]
+      exps[i].sendReset()
+      exps[i].infer(sensations, reset=False)
+
+
+    # Ensure L2 and L4 representations are consistent across all experiment
+    # instantiations, across all columns, and across 10 different repeats
+    for i in range(10):
```
mrcslws: Is this outer loop doing anything? I think it's just repeating the same thing 10 times. It walks the 10 experiments deeper in the loop.

tests/layers/l2_l4_inference_test.py
```diff
@@ -192,6 +204,139 @@ def testCapacity(self):
...
+      for c in range(5):
+        total = sum(exps[0].getL2Representations()[c])
+        for e in range(numExps):
+          self.assertEqual(total, sum(exps[e].getL2Representations()[c]))
+
+        total = sum(exps[0].getL4Representations()[c])
+        for e in range(numExps):
+          self.assertEqual(total, sum(exps[e].getL4Representations()[c]))
```
mrcslws: Would it make sense to use `set` instead of `sum`? `assertEqual` prints a useful comment when two sets aren't equal. It's harder to make sense of two unequal sums.

mrcslws:
Regarding the test failure, it's happening because the network API is calling the regions in a different order in different trials. e.g. in one trial the order is

~~~
L2Column_3
L2Column_4
L2Column_0
L2Column_1
L2Column_2
~~~

but then in another it's

~~~
L2Column_0
L2Column_1
L2Column_2
L2Column_3
L2Column_4
~~~

Since the links aren't delayed, this causes differences.

The test passes if you give the columns an extra timestep to resolve, i.e. by appending `* 2` to the `sensations`.

tests/layers/l2_l4_inference_test.py
subutai: It was there to ensure repeated accesses to the same buffer gives consistent results. Luiz at one point thought he was seeing an error there. I'll change it to two - 10 is overkill.

tests/layers/l2_l4_inference_test.py
subutai: thanks - fixed

PR #736, opened 2017-04-09
Add a "single-layer location" experiment

mrcslws:
~~This PR has two parts.~~

~~Part 1: "addMonitor" and "removeMonitor" methods`~

~~I added the notion of a "monitor" to the ColumnPooler and to two TM classes. This allowed me to separate experiment code from logging code.~~

~~This is a small change, but it's worth discussing.~~

~~This accomplishes something similar to a MonitorMixin, but it's better in a few ways:~~

- ~~This approach works with C++ classes. We already use this approach [in the Connections](https://github.com/numenta/nupic.core/blob/c117fbe3aa1b0ad49607a6854835f7351b954ef7/src/nupic/algorithms/Connections.hpp#L539). If you try to use a mixin on a SWIG'd class, your overrides won't get called when C++ code calls the method.~~
- ~~You don't have to declare a new class to use it. You can subscribe and unsubscribe on an already-existing instance. You don't have to choose between "visualizable" and "nonvisualizable" algorithms.~~
- ~~Mixins treat two things as one. For example, a single MonitoredTemporalMemory object is both a TemporalMemory and a Monitor -- two conceptually very different things. This approach explicitly treats the monitor and the algorithm class as two different things. This avoids issues where, for example, code expects to be given a MonitoredTemporalMemory and is instead given a raw TemporalMemory.~~
- ~~It doesn't require you to design extensibility into the algorithm's methods. You can provide any events you want, and you can write your class the way you want. The two are independent.~~

## Part 2: A single-layer location layer, experiment, and visualization

I added code in two places:

- `projects/location_layer`
  - A `visualizations` folder which contains JavaScript and an npm package.json for fetching dependencies and outputting a single js file, and a python package for printing these diagrams in a notebook. I decided it was best to keep this visualization code close to the experiments -- in general, each experiment will have some dedicated JavaScript code.
  - A `single_layer_2d_experiment` folder which contains a couple experiments and the logging logic for the visualization.
- `htmresearch/algorithms`
  - `single_layer_location_memory.py`. I could move that to the project folder if that'd be better.

I added two pages to the gh-pages branch:

- The main writeup: http://numenta.github.io/nupic.research/documents/location-layer/infer-2d-objects.html
- A visualizer for these CSV files: http://numenta.github.io/nupic.research/visualizations/location-layer/single-layer-2d.html

My visualization code is all in one js file and it's not very readable. Also, it assumes you're pretty familiar with d3. If/when I make more of these diagrams, I'll clean up the code. (For this PR I focused more on doing a good job on the algorithm, the experiment, the visualization itself, and the writeup.)


subutai:
The monitor approach seems unnecessarily clunky to me - whole bunch of extra code in each algorithm file that is not really tied to the algorithm.  

The monitor implementations are very specific to each experiment.  Since the runner is explicitly calling compute, why can't it just extract the information it needs after the call?  That would be equivalent to afterCompute, much simpler and avoid polluting the algorithm code.

mrcslws:
You're right. I got rid of the monitors that were in algorithm code. I still use one in the experiment code to keep it separate from the CSV writer code.

Thanks, I needed the fresh pair of eyes. I added the monitors in an early version of this experiment because at the time it seemed like the cleanest solution. But over time I continued cleaning up the experiment, and eventually the monitors became much less necessary, and I just didn't notice.

PR #738, opened 2017-04-10
Res 507 - Add basic test script

subutai:


subutai:
@mrcslws I think I made the tests deterministic now

mrcslws:
Is there a reason we don't just use folders for each algorithm? `column_pooler`, `apical_tiebreak_temporal_memory`, `apical_dependent_temporal_memory`?

It's hard to make sense of why there are `apical_tiebreak_sequences_test.py` and `apical_tiebreak_sequence_memory_test.py` files. I mean, I know the answer is that there are two underlying test suites: "apical tiebreak sequences test" and "sequence memory test", and you avoided calling something `apical_tiebreak_apical_tiebreak_sequences_test`. That makes sense, but I think the end result is still confusing.

With the previous naming, most of the filename was devoted to the name of the test, and the implementation was just prepended as a quick abbreviation (etm, attm, adtm), and the folder name expanded this abbreviation. I think that was clearer?

subutai:
Originally there was only one file per directory which looked weird, but I see your point.  I'll put it back the way it was. I also want the top level of tests to mirror htmresearch.  See what you think about the latest version!

PR #739, opened 2017-04-11
[RES-511] Fix 2 tests that are related to the temporal pooler

ywcui1990:
@subutai 
Fix  logistic_excite_function_test.py
Fix regions/temporal_pooler_region_test.py


subutai:
Thanks. While you are at it, can you also fix union_pooler_test.py? I think import should be:
`from htmresearch.algorithms.union_temporal_pooler import UnionTemporalPooler`

ywcui1990:
Sure, I will take a look at the union_pooler_test as well

PR #744, opened 2017-04-13
Experiment: Make L4L2Experiment use raw algorithms

mrcslws:
This is dependent on https://github.com/numenta/nupic.research.core/pull/14. Also, I'll need to update the L2L4 delayed links unit tests before merging (if we decide to merge this).

I made this change so that I could measure the performance overhead of the Network API, and so that I could see how clean L4L2 code is afterward. I'm using this PR to post the results, and to see whether we want to use this approach.

## The experiment

From `multi_column_convergence.py` I ran:

~~~python
    featureRange = [5]
    objectRange = [2,5,10,50]

    for numFeatures in featureRange:
      for numObjects in objectRange:
        runExperiment({
            "numObjects": numObjects,
            "numPoints": 10,
            "numLocations": 10,
            "numFeatures": numFeatures,
            "numColumns": numColumns,
            "trialNum": 4,
            "pointRange": 1,
            "plotInferenceStats": False,  # Outputs detailed graphs
            "settlingTime": 3,
            "includeRandomLocation": False,
        })
~~~

I ran it with `numColumns` 1, 3, and 8. I ran it before and after this change. I confirmed that the results are identical, then I measured performance. For each of these column counts, I ran the experiment twice on the "Before" and twice on the "After", and compared the averages.

For perf testing I made the following tweaks:

- Disabled the `updateInferenceStats` method in `l2_l4_inference.py` because it's faster in my code in a way that's sort of unfair.
- Set "checkInputs" in L4 to False.
  - I also included this change in this PR. My point is that I made a similar change to the old code when comparing performance.

### 1 Column

**Before**: 3.27 seconds
**After**: 2.35 seconds

=> Network API overhead is 28.2%

### 3 Columns

**Before**: 11.23 seconds
**After**: 7.60 seconds

=> Network API overhead is 32.3%

### 8 Columns

**Before**: 54.0 seconds
**After**: 39.9 seconds

=> Network API overhead is 26.2%


tests/algorithms/extended_temporal_memory/tm_unit_test_base.py
```diff
@@ -343,7 +344,7 @@ def testNoChangeToMatchingSegmentsInPredictedActiveColumn(self):
 
 
     tm.compute(previousActiveColumns)
-    self.assertEqual(expectedActiveCells, tm.getPredictiveCells())
+    np.testing.assert_equal(expectedActiveCells, tm.getPredictiveCells())
```
vitaly-krugl: @mrcslws, what's the advantage of `np.testing.assert_equal` in these cases?

mrcslws: This was part of the commit "ETM unit tests: don't fail on numpy arrays".  `self.assertEqual([1,2,3], np.array([1,2,3]))` will fail because only one of them is a list. `np.testing.assert_equal` will compare lists to numpy arrays without complaining.

subutai:
Cleaning up the stale PRs.

PR #746, opened 2017-04-14
Remove the Python ExtendedTemporalMemory

mrcslws:
This change:

- Deletes `extended_temporal_memory.py`
- Switches everybody to use the C++ ETM.
- Removes a few last Python ETM tests.
- Changes the ETM tests to stop using abstract methods to plug in different implementations. This makes the ETM test folder much more tidy.
  - Note that we still use abstract methods for the tests in `htmresearch/support/shared_tests`

The role of the Python ETM is now filled by:

- The Python ApicalTiebreakTemporalMemory, which essentially implements the same algorithm, but is much much faster.
- The C++ ETM
  - Still the fastest
  - Implements learnOnOneCell.


htmresearch/regions/ExtendedTMRegion.py
```diff
@@ -259,8 +259,7 @@ def getSpec(cls):
           accessMode="Read",
           dataType="Byte",
           count=0,
-          constraints=("enum: etm_py, etm_cpp, monitored_etm_py, "
-                       "monitored_etm_cpp"),
+          constraints=("enum: etm, monitored_etm, monitored_etm"),
```
subutai: There are two `monitored_etm` strings.

mrcslws: Thanks, nice catch.

PR #753, opened 2017-05-09
Run capacity experiment with different accuracy metrics

ywcui1990:
@subutai  Please review.

Summary

* Save experiment log for individual test trials (previously we only save the summary statistics for 100 trials)

* Using more stringent accuracy metrics. A test trial will be labeled as successful if 1. overlap with the correct object is the highest (no ties) 2. overlap with the non-target objects needs to be smaller than a predefined threshold. The figures below show the capacity result when this threshold is set to be 35，20, 10 and 5. The number of L2 sdr size is set to be 40.

Threshold = 35
image

Threshold = 20
image

Threshold = 10
image

Threshold = 5
image

### DIFF
--- a/projects/l2_pooling/capacity_test.py
+++ b/projects/l2_pooling/capacity_test.py
@@ -226,7 +226,7 @@ def testNetworkWithOneObject(objects, exp, testObject, numTestPoints):
 
 
 
-def testOnSingleRandomSDR(objects, exp, numRepeats=100):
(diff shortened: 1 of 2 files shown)


ywcui1990:
@subutai 

Here is the new capacity result with a threshold=30
image

And this is what we had before (note that the range of L2/L4 size is different):
image

subutai:
Great!

PR #756, opened 2017-05-19
Run thing data with l2-l4 network

ywcui1990:
@subutai 

I tried to run L4-L2 network on simulated data from Luiz. The results are shown below. In this example I wrote a loadThingObjects function and then call the "learnObjects" function of the L4L2Experiment. I felt this is more straightforward than using the ObjectMachine. What do 
you think?

image


projects/l2_pooling/thing_convergence.py
```diff
@@ -0,0 +1,112 @@
...
+def createExperiment(logFilename):
+  # Typically this would be done by Thing
+  exp = L4L2Experiment("shared_features", logCalls=True)
+  exp.learnObjects(thingObjects)
+
+  LoggingDecorator.save(exp.callLog, logFilename)
+
+
+def debugExperiment(logFile, thingObjects):
+  """
```
subutai: Not sure what is going on here - what is the logfile and why is this called debug?

projects/l2_pooling/thing_convergence.py
```diff
@@ -0,0 +1,112 @@
...
+        len(exp.getL2Representations()[0] & L2Representations[objects[i]][0]))
+    exp.sendReset()
+
+
+
+if __name__ == "__main__":
+  thingObjects = loadThingObjects()
+
+  # Mimic thing, which will create a log file
+  createExperiment("callLog.pkl")
```
subutai: ???

ywcui1990:
@subutai Thanks for the feedback. I have updated the PR significantly. Please review projects/thing_classification/thing_convergence.py when you have some time.

PR #757, opened 2017-05-23
Add capacity experiment: # of locations/features

mrcslws:
@ywcui1990 Here's the current result. I tested 50, 250, 450, and 650 objects.

image


mrcslws:
@ywcui1990 with the bugfix, here are the new results.

image

ywcui1990:
Nice!

PR #758, opened 2017-05-23
Energy based pooling

mirkoklukas:
The aim is to express Numenta’s spatial pooler as a Hopfield network. 

Everything is contained in a new project folder `htmresearch/projects/energy_based_pooling/`


ywcui1990:
@mirkoklukas Thanks for sharing this work. I will review it.

Two quick comments for now:
Can you add the AGPL license statement like [this](https://github.com/numenta/htmresearch/blob/master/projects/l2_pooling/multi_column_convergence.py#L1)?

Can you remove the video files from this PR? We don't want videos in this repo.

PR #759, opened 2017-05-25
Faster capacity test: Use SparseMatrix to calculate object overlaps

mrcslws:
@ywcui1990 A lot of time was spent computing overlaps between L2 and every object. This change eliminates almost all of that overhead.

If I run a small capacity experiment, here are the times:

**Before**: 200 seconds
**After**: 114 seconds

So it now takes ~57% of the time that it previously took.


ywcui1990:
:+1:

PR #760, opened 2017-05-25
Feedback experiment with new Apical Modulation implementation of ApicalTMRegion.py

ThomasMiconi:
We add a new implementation for ApicalTMRegion.py, called Apical Modulation.

Now active apical segments lower the threshold for basal (lateral) segment activation (in addition to breaking ties within a minicolumn, as inherited from ApicalTiebreak).

We also implement a different measure of predictive performance, namely, correlation between predicted pattern and firing pattern in the absence of noise (that is, the 'correct' firing pattern). Other measures are tracked and can be chosen for the main graph by modifying feedback_sequences.py.

Feedback_experiment doesn't use laminar_network and l4_l2_creation anymore: the network is directly created in the feedback_experiment.py module. It can use either the ExtendedTMRegion or ApicalTMRegion, with either the new ApicalModulation or the old ApicalTiebreak implementation (see line 90 in feedback_experiment.py).

Also, column_pooler (the implementation for layer 2) is computed slightly differently than before, with inertia (check lines 288-336 in column_pooler.py).


htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +280,68 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
 
-    # If necessary, activate some of the previously active cells
-    if len(chosenCells) < minNumActiveCells:
-      remaining = numpy.setdiff1d(prevActiveCells, feedforwardSupportedCells)
-      remaining = remaining[numActiveSegmentsByCell[remaining] > 0]
+    chosenCells = []
+    minNumActiveCells =  30 #self.sdrSize / 2
+
+
+    # # This is the old method. It relies heavily on the chooseCells function, which is
```
subutai: Lot of commented out code.  If the changes look reasonable and we get the same results on our convergence scripts I would rather we just remove the commented out parts.  @mrcslws

subutai:
@mrcslws I asked @ThomasMiconi to verify with our convergence scripts.

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +280,68 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+    #   chosenCells = numpy.append(
+    #     chosenCells, self._chooseCells(remaining,
+    #                                    minNumActiveCells - len(chosenCells),
+    #                                    numActiveSegmentsByCell))
+
+
+    # New method. Doesn't use the chooseCells function.
+    # # First, select FF-supported cells. But if some have active segments, only chose these ones (they inhibit everyone else)
+    remainingFFSupportedCells = numpy.setdiff1d(feedforwardSupportedCells, chosenCells)
+    activeSegs = numActiveSegmentsByCell[remainingFFSupportedCells]
```
mrcslws: I think this variable name should indicate that it's the "# of segments for feedforward supported cells".  Perhaps `numActiveSegsForFFSCells`.

It corresponds to the `remainingFFSupportedCells` was used to create it, so it's useful if their variable names indicate that.

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +280,68 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+    if len(remainingFFSupportedCells) == 0:
+        pass  # Nothing to add
+    elif numpy.max(activeSegs) == 0:
+         chosenCells = numpy.append(chosenCells, remainingFFSupportedCells)
+    else:
+         chosenCells = numpy.append(chosenCells, remainingFFSupportedCells[activeSegs > 0])
```
mrcslws: We can't just activate every cell that has an active segment. I expect this will cause bad results in the "multi_column_convergence" experiment. Basically, cortical columns won't help each other -- they'll actually keep each other from inferring the object.

htmresearch/algorithms/apical_modulation_temporal_memory.py
```diff
@@ -0,0 +1,917 @@
...
+
+    - potentialOverlaps (numpy array)
+      The number of active potential synapses for each segment.
+      Includes counts for active, matching, and nonmatching segments.
+    """
+
+    # Active apical segments lower the activation threshold for basal (lateral) segments
+    overlaps = connections.computeActivity(activeInput, connectedPermanence)
+    outrightActiveSegments = np.flatnonzero(overlaps >= activationThreshold)
+    conditionallyActiveSegments = np.flatnonzero((overlaps < activationThreshold) & (overlaps >= activationThreshold * 0.75))
```
mrcslws: I agree that this should be a threshold parameter, not a factor. My vote is that we add a new parameter, not use minThreshold. minThreshold is essentially a "learningThreshold" and probably shouldn't be tangled with this.

htmresearch/regions/ApicalTMRegion.py
```diff
@@ -265,7 +273,7 @@ def getSpec(cls):
           "accessMode": "Read",
           "dataType": "Byte",
           "count": 0,
-          "constraints": ("enum: ApicalTiebreak, ApicalDependent"),
+          "constraints": ("enum: ApicalTiebreak, ApicalDependent, ApicalModulation"),
```
mrcslws: The new one is no longer needed.

htmresearch/regions/ApicalTMRegion.py
```diff
@@ -341,6 +351,7 @@ def initialize(self):
         "apicalInputSize": self.apicalInputWidth,
         "cellsPerColumn": self.cellsPerColumn,
         "activationThreshold": self.activationThreshold,
+        "reducedThresholdBasal": self.reducedThresholdBasal,
```
mrcslws: We'll probably want to move this line down to the `self.implementation == "ApicalTiebreak"` case.

i.e.

~~~python
params["reducedThresholdBasal"] = self.reducedThresholdBasal
self._tm = ApicalTiebreakTemporalMemory(**params)
~~~

htmresearch/regions/ApicalTMRegion.py
```diff
@@ -287,6 +295,7 @@ def __init__(self,
                initialPermanence=0.21,
                connectedPermanence=0.50,
                minThreshold=10,
+               reducedThresholdBasal=11, # Only used for apicalTiebreak implementation
```
mrcslws: We should use the same default value in the `ApicalTMRegion` as in the `ApicalTiebreakTemporalMemory`.

htmresearch/regions/ApicalTMRegion.py
```diff
@@ -357,6 +368,10 @@ def initialize(self):
         from htmresearch.algorithms.apical_tiebreak_temporal_memory import (
           ApicalTiebreakTemporalMemory)
         self._tm = ApicalTiebreakTemporalMemory(**params)
+      elif self.implementation == "ApicalModulation":
```
mrcslws: This can go. (As you know)

projects/feedback/feedback_experiment.py
```diff
@@ -289,7 +390,14 @@ def getL4PredictiveCells(self):
     """
     Returns the predictive cells in L4.
     """
-    return [set(column._tm.getPredictiveCells()) for column in self.L4Columns]
+    # ApicalTMRegion uses "getPredictedCells", while ExtendedTMRegion uses "getPredictiveCells".
+    #return [set(column._tm.getPredictiveCells()) for column in self.L4Columns]
+    if self.L4RegionType == "py.ApicalTMRegion":
+        return [set(column._tm.getPredictedCells()) for column in self.L4Columns]
```
mrcslws: In this PR you're using a lot of 4-space tabs. Can you convert these all to 2-spaces-per-tab? It's a little messy seeing a mix of 4 and 2.

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -475,6 +489,56 @@ def _calculateSegmentActivity(connections, activeInput, connectedPermanence,
             matchingSegments,
             potentialOverlaps)
 
+  @staticmethod
```
mrcslws: Nitpick: put two newlines between each method.

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -475,6 +489,56 @@ def _calculateSegmentActivity(connections, activeInput, connectedPermanence,
...
+    """
+    # Active apical segments lower the activation threshold for basal (lateral) segments
+    overlaps = connections.computeActivity(activeInput, connectedPermanence)
+    outrightActiveSegments = np.flatnonzero(overlaps >= activationThreshold)
+    if reducedThresholdBasal != activationThreshold:
+        conditionallyActiveSegments = np.flatnonzero((overlaps < activationThreshold)
+                                        & (overlaps >= reducedThresholdBasal))
+        cellsOfCASegments = connections.mapSegmentsToCells(conditionallyActiveSegments)
+        # apically active segments are condit. active segments from apically active cells
+        apicallyActiveSegments = conditionallyActiveSegments[np.in1d(cellsOfCASegments,
```
mrcslws: I actually added a `filterSegmentsByCell` method that will be useful here:

~~~python
apicallyActiveSegments  = connections.filterSegmentsByCell(
  conditionallyActiveSegments, apicallyActiveCells)
~~~

mrcslws:
I spent some time thinking through the pooling layer algorithm.

Fundamentally, I think your change takes these two groups:

- Cells with feedforward support and no lateral support
- Cells that were previously active

and it reprioritizes them. The current algorithm favors the first group over the second. Your algorithm activates some of each, favoring the second group slightly more than the first.

All-in-all I think I like this "inertia" approach.

Here are some issues that stand out.

## This only helps when at least one cell has lateral support

When no cells have lateral support, your algorithm is essentially identical to the current algorithm. You activate every cell with feedforward support, then maybe fall back to previously active cells.

When some cells have lateral support, that's when your algorithm becomes different. The cells with lateral support are all activated, then we mainly fall back to the previously active cells rather than the feedforward supported cells.

I think you could take your approach further. Your principle of favoring previously active cells is a good one. I think you should start by activating a big portion of previously active cells, **then** activate some of the cells with feedforward support. At the bottom of this write-up I showed one way you could structure this. This change wouldn't affect our noise-free object inference experiments. Our experiments never activate cells that weren't previously active.

## Sudden behavior changes between 0 and 1

Imagine 120 cells have feedforward support, and 0 have lateral support. _(This happens a lot in our object inference models.)_ All 120 cells will activate.

Now imagine 120 cells have feedforward support, and 1 has lateral support. With our current algorithm, 120 cells will activate. With this change, 30 will activate.

I think we want to avoid these sudden changes in outcomes at the boundaries. It will cause quirky behavior. I think the correct approach here is: look at the two boundaries, 1 and 39, and make sure the right thing happens at each boundary. Then somehow interpolate between those two outcomes. In the current algorithm, we chose an abrupt change of behavior at 20, but I think it would have been better if we'd used smoother interpolation.

In my [fault tolerant L2 notebook](https://github.com/numenta/htmresearch/blob/794b4b4b26c33ac723867c07d8bbd78ff037b2e8/projects/l2_pooling/notebooks/fault_tolerant_L2.ipynb) I described one way we might do this interpolation:

> With this change, cells are inhibited proportionally to the number of already-activated cells. The further we are from 40 active cells, the more of the next group that we'll activate. So if we have 10 active cells so far, there's a discrepancy of 30 cells, so we will activate 3/4 of the next group. (Though we'll make sure to activate at least 30, if possible.)

Anyway, I think those are the options. We shouldn't have dramatic differences between 0<->1 and 39<->40. We can solve this by smooth interpolation, or by having an abrupt threshold in the middle.

## Weird behavior with segment counts

In the current algorithm, cells with 3 active segments are favored over cells with 2, which are favored over cells with 1, and so on. With your change, this is mostly true, but it's quirky.

Imagine the feedforward-supported cells consist of:

- 10 with 3 active segments
  - all previously active
- 20 with 2 active segments
  - all previously active
- 100 with 0 active segments

Your algorithm will activate:

- 10 with 3 active segments
- 3 random cells of the 120 cells with < 3 active segments
- ~17 with 2 active segments

This probably works. But it seems a little bit unprincipled. I can't really give a straight explanation of why some of the "100" got activated.

When we're activating cells with feedforward support, I think it would be better to favor the ones with more active segments, possibly regulating it with the "distalSegmentInhibitionFactor".

## Precise segment counting, not configurable

This change removes the "distalSegmentInhibitionFactor". That means that, with this change, cells with 6 active segments will inhibit cells with 5 active segments.

We didn't add this factor to improve results. We added it because it made everything more biologically plausible. Neurons probably aren't doing precise segment counting. We gave ourselves a parameter to tweak the precision of the segment counting. When this factor is 1.5, that means a cell with 3 active segments will inhibit a cell with 2, but a cell with 4 won't inhibit a cell with 3.

Recently we've mostly disabled this feature. We set this factor close to 1.0, enabling precise segment counting. So it might be fine to remove it. But it would be nice if we could consider that detail independently of this PR -- i.e. if we could find a way to have a distalSegmentInhibitionFactor or some similar parameter with this PR.

## How I might change this

I like this approach. I think you've applied this approach really well at the boundary between 0 and 1 active segments. Here's how I think I'd apply it to the other boundaries.

I think I would prioritize the groups as follows. I've rearranged your logic a bit: you activate 1/6 of the remainder with cells with feedforward support, then fall back to the prev active cells. I flipped it around. I activate 5/6 (or ~80%) of the prev active cells, then I fall back to the cells with feedforward support. _(For this example, I arbitrarily chose 3 as the max number of active segments.)_

Priorities:

1. Previous active + >=3 active segments
    - Activate 80% of remainder (or five-sixths of 30, using your thresholds)
2. Feedforward support + 3 active segments
3. Previous active + >=2 active segments
    - Activate 80% of remainder
4. Feedforward support + 2 active segments
5. Previous active + >=1 active segments
    - Activate 80% of remainder
6. Feedforward support + 1 active segments
7. Previous active + >=0 active segments
    - Activate 80% of remainder
8. Feedforward support + 0 active segments
9. **Maybe** activate more previous active if we still haven't reached the minimum

I would activate groups in this order. I'd probably regulate it with the distalSegmentInhibitionFactor. And I'd do some clever logic to figure out exactly how much of a group to activate. (Trying to avoid abrupt changes at the boundaries).

Note the `>=` signs. That's important to avoid some weird situations, I think.

I like the idea of skipping Group 9. If no cells in the layer have feedforward support, the number of active cells would decay every time step. (It's goofy that our current algorithm, with no feedforward suppport, will keep cells active indefinitely if they're providing lateral support to each other).

______

Hopefully this was useful! Definitely let me know if I'm overlooking anything. (I probably am.)

ThomasMiconi:
OK

I have implemented the requested changes for the Apical Tie Break temporal memory (changed parameter names, removed unneeded elements, etc.)

I left the column pooler unchanged for now. The proposed modifications are more consequent than I feel comfortable with at the moment.

In addition, I have also added new changes to selectively disable individual components of the feedback process: inertia, apical tie-break, and apical modulation of the basal threshold).

These latter changes are used in the new file feedback_sequences_additional.py, which runs experiments with separate feedback components disabled in order to compare their impacts.

mrcslws:
This is a pretty big change to the column pooler. In my comment above, I listed some issues in this change:

- Toggling a segment on a single cell can cause a sudden dramatic change in the entire layer's behavior. So a lateral false negative will have a larger effect.
- When different cells have different amounts of segments active, the outcome is kinda weird. I can't really explain why having a mix of cells with 2-3 active segments would cause some cells with 0 active segments to activate.
- It uses precise segment counting, removing the "distalSegmentInhibitionFactor" which controls the precision of the counting. This parameter is useful because it might make our model more biologically plausible.

I think you're more focused on single-cortical-column noise tolerance. For now, maybe you should not use the column pooler, and instead create a new class? And then in a future change we could figure out the right way to fit these principles into the column pooler, and in that change we'd focus on multi-column testing. (We could discuss offline)

mrcslws:
I think this is a very reasonable change. The algorithm is ready to be merged, after some code cleanup.

Here's how I think of this change. There are two ways to add noise tolerance:

1. Make it harder for noise to inactivate active cells. Keep many previous active cells active.
2. Make it harder for noise to activate new cells. Keep many previous inactive cells inactive.

The latest version of your change does 1 and not 2. As you've mentioned, we could further improve performance by also doing 2 . Experimenting with this PR, I've observed that 1 has more impact on the noise tolerance of sequence memory, while 2 has more impact on the noise tolerance of object inference. In the future we might want to also do 2, but we don't need to do that here, especially since these experiments are focused on noise tolerance of sequence memory.

Let me know when you're ready for me to code-review this. (There's a lot of commented code, 4-space-tabs, unnecessary calls to numpy.permutations, >80 character lines. Also, feel free to "import numpy as np" in `column_pooler.py` if it will help reach 80 characters per line.)

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,198 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+    else:
+        # This loop will select the FF-supported AND laterally-active cells, in
+        # order of descending lateral activation, until we exceed the
+        # minNumActiveCells quorum - but will exclude cells with 0 lateral
+        # active segments!
+        ttop = numpy.max(numActiveSegsForFFSuppCells)
+        if ttop > 0:
+            while True:
+                chosenCells = numpy.unique(numpy.append(chosenCells,
+                        remainingFFSupportedCells[numActiveSegsForFFSuppCells >= self.distalSegmentInhibitionFactor * ttop]))
```
mrcslws: This should probably be `>`, not `>=`. For example, if this factor is `.5`, and if `ttop` is 2, then we probably want cells with 1 segment to be inhibited.

(I like that you changed the `distalSegmentInhibitionFactor` to its inverse. This math is much clearer.)

htmresearch/algorithms/column_pooler.py
```diff
@@ -53,7 +53,8 @@ def __init__(self,
                sampleSizeDistal=20,
                activationThresholdDistal=13,
                connectedPermanenceDistal=0.50,
-               distalSegmentInhibitionFactor=1.001,
+               distalSegmentInhibitionFactor=0.999,  #Should be <1
```
mrcslws: Instead of putting "Should be <1" here, we should probably:

1. Update the docstring for distalSegmentInhibitionFactor below.
2. Change the `assert distalSegmentInhibitionFactor > 0` to `assert distalSegmentInhibionFactor < 1`.

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
         lateralInput, self.connectedPermanenceDistal)
       numActiveSegmentsByCell[overlaps >= self.activationThresholdDistal] += 1
 
-    # Activate some of the feedforward supported cells
-    minNumActiveCells = self.sdrSize / 2
-    chosenCells = self._chooseCells(feedforwardSupportedCells,
-                                    minNumActiveCells, numActiveSegmentsByCell)
+    chosenCells = []
+    minNumActiveCells =  30 #self.sdrSize / 2
```
mrcslws: This 30 should either be a parameter, or it should be `self.sdrSize * 0.75`. (I'm assuming you found that 30 was better than 20.)

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
 
-    # If necessary, activate some of the previously active cells
-    if len(chosenCells) < minNumActiveCells:
-      remaining = numpy.setdiff1d(prevActiveCells, feedforwardSupportedCells)
-      remaining = remaining[numActiveSegmentsByCell[remaining] > 0]
+    remainingFFSupportedCells = numpy.setdiff1d(feedforwardSupportedCells,
+                                                    chosenCells)
```
mrcslws: This line isn't doing anything. `chosenCells` is empty here.

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+    numActiveSegsForFFSuppCells=numActiveSegmentsByCell[remainingFFSupportedCells]
```
mrcslws: nitpick: put a space on either side of the `=`

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
 
-      chosenCells = numpy.append(
-        chosenCells, self._chooseCells(remaining,
-                                       minNumActiveCells - len(chosenCells),
-                                       numActiveSegmentsByCell))
+    # First, activate the FF-supported cells that have the highest number of
+    # lateral active segments (this may be 0)
```
mrcslws: "this may be 0" is no longer true.

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+    if len(remainingFFSupportedCells) == 0:
+        pass
```
mrcslws: nitpick: 2 spaces, not 4

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+      if self.useInertia:
+          chosenCells = numpy.append(chosenCells,
+                prevCells[:int(len(prevCells) * self.inertiaFactor)] )
 
     chosenCells.sort()
     self.activeCells = numpy.asarray(chosenCells, dtype="uint32")
 
 
+  # chooseCells is not used in the current version, but preserved here.
   def _chooseCells(self, candidates, n, numActiveSegmentsByCell):
```
mrcslws: We don't need to preserve `_chooseCells`. You can remove it.

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+      # active segments!
+      ttop = numpy.max(numActiveSegsForFFSuppCells)
+      if ttop > 0:
+        while True:
+          chosenCells = numpy.unique(numpy.append(chosenCells,
+                    remainingFFSupportedCells[numActiveSegsForFFSuppCells >
+                    self.distalSegmentInhibitionFactor * ttop]))
+          ttop -= 1
+          if ttop <= 0 or len(chosenCells) > minNumActiveCells:
+              break
```
mrcslws: You could remove lines 302, 303, 308, 309, and just do 

~~~python
while ttop > 0 and len(chosenCells) > minNumActiveCells:
  ...
~~~

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+      # This loop will select the FF-supported AND laterally-active cells, in
+      # order of descending lateral activation, until we exceed the
+      # minNumActiveCells quorum - but will exclude cells with 0 lateral
+      # active segments!
+      ttop = numpy.max(numActiveSegsForFFSuppCells)
+      if ttop > 0:
+        while True:
+          chosenCells = numpy.unique(numpy.append(chosenCells,
+                    remainingFFSupportedCells[numActiveSegsForFFSuppCells >
+                    self.distalSegmentInhibitionFactor * ttop]))
```
mrcslws: Instead of doing `append` + `unique`, you can just do 

~~~python
chosenCells = numpy.union1d(chosenCells, otherCells)
~~~

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+          ttop -= 1
+          if ttop <= 0 or len(chosenCells) > minNumActiveCells:
+              break
+
+    # If we still haven't filled the minNumActiveCells quorum, add in the
+    # FF-supported cells with 0 lateral support AND the inertia cells.
+    if len(chosenCells) < minNumActiveCells:
+      prevCells = numpy.setdiff1d(prevActiveCells, chosenCells)
+      numActiveSegsForPrevCells = numActiveSegmentsByCell[prevCells]
+      # We sort the previously-active cells by nuber of active lateral
```
mrcslws: Typo: "nuber"

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+      # segments (this really helps)
+      prevCells = prevCells[numpy.argsort(numActiveSegsForPrevCells)[::-1]]
+      remFFcells = numpy.setdiff1d(remainingFFSupportedCells, chosenCells)
+      # Note that this is all the remaining FF-supported cells!
+      chosenCells=numpy.append(chosenCells,remFFcells)
```
mrcslws: nitpick: put a space on both sides of the '='

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+      prevCells = numpy.setdiff1d(prevActiveCells, chosenCells)
+      numActiveSegsForPrevCells = numActiveSegmentsByCell[prevCells]
+      # We sort the previously-active cells by nuber of active lateral
+      # segments (this really helps)
+      prevCells = prevCells[numpy.argsort(numActiveSegsForPrevCells)[::-1]]
+      remFFcells = numpy.setdiff1d(remainingFFSupportedCells, chosenCells)
+      # Note that this is all the remaining FF-supported cells!
+      chosenCells=numpy.append(chosenCells,remFFcells) 
+      if self.useInertia:
+          chosenCells = numpy.append(chosenCells,
```
mrcslws: 2 spaces, not 4

htmresearch/algorithms/column_pooler.py
```diff
@@ -273,25 +282,52 @@ def _computeInferenceMode(self, feedforwardInput, lateralInputs):
...
+                prevCells[:int(len(prevCells) * self.inertiaFactor)] )
```
mrcslws: When `self.useInertia` is False, we're unnecessarily computing `prevCells` and `numActiveSegsForPrevCells`. It'd be good to do those after the `if self.useInertia:`.

htmresearch/algorithms/column_pooler.py
```diff
@@ -491,6 +527,19 @@ def reset(self):
     """
     self.activeCells = numpy.empty(0, dtype="uint32")
 
+  def getUseInertia(self):
+        """
+        Get whether we actually use apical modulation of basal threshold.
+        @return (Bool) Whether apical modulation is used.
```
mrcslws: This docstring is not for `useInertia`. Also, use 2 spaces, not 6.

htmresearch/algorithms/column_pooler.py
```diff
@@ -491,6 +527,19 @@ def reset(self):
...
+        """
+        return self.useInertia
+
+  def setUseInertia(self, useInertia):
+            """
+            Sets whether we actually use inertia.
+            @param useInertia (Bool) Whether inertia is used.
+            """
+            self.useInertia = useInertia
```
mrcslws: 2 spaces

htmresearch/regions/ColumnPoolerRegion.py
```diff
@@ -291,6 +300,7 @@ def __init__(self,
                activationThresholdDistal=13,
                connectedPermanenceDistal=0.50,
                distalSegmentInhibitionFactor=1.001,
```
mrcslws: We need to update this factor's default value. You'll probably want to search the htmresearch repo and update all values. https://github.com/numenta/htmresearch/search?utf8=✓&q=distalSegmentInhibitionFactor&type=

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -201,18 +212,26 @@ def compute(self,
...
-     apicalPotentialOverlaps) = self._calculateSegmentActivity(
+     apicalPotentialOverlaps) = self._calculateApicalSegmentActivity(
        self.apicalConnections, apicalInput, self.connectedPermanence,
        self.activationThreshold, self.minThreshold)
 
+    if learn or self.useApicalModulationBasalThreshold==False:
+        reducedBasalThresholdCells = ()
+    else:
+        reducedBasalThresholdCells = self.apicalConnections.mapSegmentsToCells(
+            activeApicalSegments)
```
mrcslws: nitpick: use 2 spaces, not 4.

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -760,6 +834,22 @@ def setActivationThreshold(self, activationThreshold):
...
+    @return (int) The activation threshold.
+    """
+    return self.reducedBasalThreshold
+
+
+  def setreducedBasalThreshold(self, reducedBasalThreshold):
+    """
+    Sets the activation threshold.
+    @param activationThreshold (int) activation threshold.
+    """
```
mrcslws: The docstring needs updating.

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -760,6 +834,22 @@ def setActivationThreshold(self, activationThreshold):
     self.activationThreshold = activationThreshold
 
 
+  def getreducedBasalThreshold(self):
+    """
+    Returns the activation threshold.
+    @return (int) The activation threshold.
+    """
```
mrcslws: The docstring needs updating.

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -760,6 +834,22 @@ def setActivationThreshold(self, activationThreshold):
     self.activationThreshold = activationThreshold
 
 
+  def getreducedBasalThreshold(self):
```
mrcslws: Capitalize the 'r'. `getReducedBasalThreshold`

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -760,6 +834,22 @@ def setActivationThreshold(self, activationThreshold):
...
+    """
+    Returns the activation threshold.
+    @return (int) The activation threshold.
+    """
+    return self.reducedBasalThreshold
+
+
+  def setreducedBasalThreshold(self, reducedBasalThreshold):
```
mrcslws: Capitalize the 'r'. `setReducedBasalThreshold `

htmresearch/algorithms/apical_tiebreak_temporal_memory.py
```diff
@@ -886,3 +976,32 @@ def setConnectedPermanence(self, connectedPermanence):
...
     """
     self.connectedPermanence = connectedPermanence
+
+  def getUseApicalTieBreak(self):
+    """
+    Get whether we actually use apical tie-break.
+    @return (Bool) Whether apical tie-break is used.
+    """
+    return self.useApicalTiebreak
+
```
mrcslws: Put two newlines between methods, not just 1.

MaxASchwarzer:
As Thomas is now gone, I'm going to be taking over this issue.  As such, I'm closing this PR, and creating a new one from my branch.  Please see https://github.com/numenta/htmresearch/pull/760 for all subsequent work.

PR #761, opened 2017-05-28
Switch to the updated ETM API

mrcslws:
Depends on https://github.com/numenta/htmresearch-core/pull/19

The ETM now uses the same API as the `ApicalTiebreakTemporalMemory`. It has no notion of "internal connections -- it receives its entire `basalInput` as a single list. And `getPredictedCells` returns predictions for the current timestep, not the next timestep.

Now `l2_l4_network_creation` detects the "formInternalBasalConnections" parameter and adds links accordingly.

I removed `tm_unit_test.py` because we already have that unit test written in C++. It made more sense when we also had a Python ETM, but now it's not worth maintaining two identical sets of tests.


mrcslws:
I backpedalled slightly and added a `basalInputPrepend` parameter to the ETM, but only in the ETM bindings. The C++ is oblivious of it. This makes life easier for people using the ETM directly who want internal + external basal connections. Still, with this PR, the C++ no longer keeps the internal and external inputs separate, so there's a small performance impact in concatenating them. (Our main experiments don't use combined internal + external connections, so it's not worth making this feature a first-class citizen of the ETM.)

With the network API, we now do sequence memory with the ETM by linking the region's `activeCells` and `winnerCells` to its own `basalInput` and `basalGrowthCandidates` with `propagationDelay=1`. (Thomas is already using this approach with the ApicalTMRegion.)

@subutai This is ready for review.

tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -834,23 +830,23 @@ def generatePattern(self, max, size):
     return cellsIndices[:size]
 
 
-  def getL4PredictiveCells(self, column):
-    """Returns the predictive cells in L4."""
-    return set(column._tm.getPredictiveCells())
+  def getL4PredictedCells(self, column):
+    """Returns the predicted cells in L4."""
```
subutai: Could you specify clearly what the term "predictedCells" means? e.g. are these the cells that have active segments as a result of the last call to compute?

PR #763, opened 2017-06-05
RES-530 Add tests to check L2L4 network link creation

MaxASchwarzer:
Add two tests to check link formation in both simple and multicolumn L2L4 networks.

### DIFF
--- a/tests/frameworks/layers/l2l4_network_creation_test.py
+++ b/tests/frameworks/layers/l2l4_network_creation_test.py
@@ -193,6 +193,52 @@ class LaminarNetworkTest(unittest.TestCase):
     net.run(3)
 
 
+  def testL4L2ColumnLinks(self):
+    """
+    In this simplistic test we create a network and ensure that it has the
+    correct links between regions.
+    """
+
+    # Create a simple network to check its architecture
+    net = createNetwork(networkConfig1)
+
+    links = net.getLinks()
+
+    # Make sure that we have the right number before going on to specifics
+    self.assertEqual(len(list(net.getLinks())), 6, "Incorrect number of links")
+
+    # These are all the links we're hoping to find
+    desired_links=[("sensorInput_0", "dataOut", "L4Column_0", "activeColumns"),
+      ("L2Column_0", "feedForwardOutput", "L4Column_0", "apicalInput"),
+      ("externalInput_0", "dataOut", "L4Column_0", "basalInput"),
+      ("L4Column_0", "predictedActiveCells", "L2Column_0",
+       "feedforwardGrowthCandidates"),
+      ("L4Column_0", "activeCells", "L2Column_0", "feedforwardInput"),
+      ("sensorInput_0", "resetOut", "L2Column_0", "resetIn")]
+
+    # This gets the actual C-based link objects and discards their tags
+    links = map(lambda x: x.second, links)
+
+    # Form the links into a text-based representation
+    links = map(lambda x: (x.getSrcRegionName(),
+    x.getSrcOutputName(), x.getDestRegionName(), x.getDestInputName()), links)
+
+    # Build a descriptive error message to pass to the user
+    error_message = "Links incorrectly formed in simple L2L4 network: \n"
+    for link in desired_links:
+      if not link in links:
+        error_message += "Failed to find link: {}.{} to {}.{} \n".format(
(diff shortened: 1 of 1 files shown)


tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -193,6 +193,52 @@ def testL4L2ColumnCreate(self):
...
+    desired_links=[("sensorInput_0", "dataOut", "L4Column_0", "activeColumns"),
+      ("L2Column_0", "feedForwardOutput", "L4Column_0", "apicalInput"),
+      ("externalInput_0", "dataOut", "L4Column_0", "basalInput"),
+      ("L4Column_0", "predictedActiveCells", "L2Column_0",
+       "feedforwardGrowthCandidates"),
+      ("L4Column_0", "activeCells", "L2Column_0", "feedforwardInput"),
+      ("sensorInput_0", "resetOut", "L2Column_0", "resetIn")]
+
+    # This gets the actual C-based link objects and discards their tags
+    links = map(lambda x: x.second, links)
```
scottpurdy: Use a list comprehension instead of `map`. `map` should pretty much never be used in Python

tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -193,6 +193,52 @@ def testL4L2ColumnCreate(self):
...
+    # Create a simple network to check its architecture
+    net = createNetwork(networkConfig1)
+
+    links = net.getLinks()
+
+    # Make sure that we have the right number before going on to specifics
+    self.assertEqual(len(list(net.getLinks())), 6, "Incorrect number of links")
+
+    # These are all the links we're hoping to find
+    desired_links=[("sensorInput_0", "dataOut", "L4Column_0", "activeColumns"),
```
scottpurdy: This heavily relies on implementation details for constructing the region "names". Is there an easy way to get the actual name as opposed to the derived one (with "_0")? That would be a better test.

You might also make this a set of tuples and simply check for set equality below. For the debug message you can just print both full lists since there aren't too many links to manually inspect.

tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -193,6 +193,52 @@ def testL4L2ColumnCreate(self):
...
+      if not link in links:
+        error_message += "Failed to find link: {}.{} to {}.{} \n".format(
+          link[0],link[1],link[2],link[3])
+
+    for link in links:
+      if not link in desired_links:
+        error_message += "Found unexpected link: {}.{} to {}.{} \n".format(
+          link[0],link[1],link[2],link[3])
+
+    self.assertEqual(sorted(desired_links), sorted(links),
```
scottpurdy: Better to us `assertSequenceEqual` when comparing two lists. And if you go the set approach, this becomes something like:

```
message = "Actual links:\n\t{}\ndon't match expected:\n\t{}".format(links, desired_links)"
self.assertSetEqual(desired_links, links, message)
```

If the message format doesn't look good then you can construct something nicer but I'm guessing this will be sufficient.

tests/frameworks/layers/l2l4_network_creation_test.py
MaxASchwarzer: Done.

tests/frameworks/layers/l2l4_network_creation_test.py
MaxASchwarzer: You're correct that we're very dependent on the region names.  Unfortunately, so far as I can tell, names like L4Column_0 _are_ the real names of the regions, at least so far as the system stores.  As a result, our other tests are equally dependent on this naming scheme. 

I'm somewhat concerned about just printing the lists, since in the more complicated version there are roughly 50 links.  Moreover, the number of links will be noticeably higher if this test scheme is extended for networks with 2D topology.

tests/frameworks/layers/l2l4_network_creation_test.py
MaxASchwarzer: Done, although I've kept the more detailed message format for the reasons stated above.

MaxASchwarzer:
Will do.  I believe the links listed in the tests are correct, but I'll email him to confirm.

PR #764, opened 2017-06-06
RES-363 2D Column Topology

MaxASchwarzer:
Add basic implementation for 2D column topology, with corresponding unit tests.


tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -281,37 +321,168 @@ def testMultipleL4L2ColumnLinks(self):
...
+    externalInput2.addDataToQueue([2, 42, 1023], 0, 9)
+    sensorInput2.addDataToQueue([2, 42, 1023], 0, 0)
+    externalInput3.addDataToQueue([2, 42, 1023], 0, 9)
+    sensorInput3.addDataToQueue([2, 42, 1023], 0, 0)
+    externalInput4.addDataToQueue([2, 42, 1023], 0, 9)
+    sensorInput4.addDataToQueue([2, 42, 1023], 0, 0)
+
+    # Run the network and check outputs are as expected
+
+    #import pdb; pdb.set_trace()
```
scottpurdy: remove?

tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -281,37 +321,168 @@ def testMultipleL4L2ColumnLinks(self):
...
+
+
+    # Spotcheck some of the phases
+    self.assertEqual(net.getPhases("externalInput_0"),(0,),
+                     "Incorrect phase externalInput_0")
+    self.assertEqual(net.getPhases("externalInput_1"),(0,),
+                     "Incorrect phase for externalInput_1")
+    self.assertEqual(net.getPhases("L4Column_0"),(2,),
+                     "Incorrect phase for L4Column_0")
+    self.assertEqual(net.getPhases("L4Column_1"),(2,),
```
scottpurdy: put a space after the commas that are separating arguments to `getPhases`

tests/frameworks/layers/l2l4_network_creation_test.py
```diff
@@ -625,7 +795,8 @@ def testTwoColumnsL4L2DataFlow(self):
 
     self.assertEqual(
       len(net.regions.keys()), 4 * 2,
-      "Incorrect number of regions"
+      "Incorrect number of regions, expected {} but had {}".format(8*2,
+      len(net.regions.keys()))
```
scottpurdy: indent since you are wrapping in the middle of `format()`, not `assertEqual()` like the previous line

htmresearch/frameworks/layers/l2_l4_network_creation.py
```diff
@@ -333,3 +359,74 @@ def createMultipleL4L2Columns(network, networkConfig):
...
+
+      suffix = "_" + str(i)
+      network = createL4L2Column(network, networkConfigCopy, suffix)
+
+    # Now connect the L2 columns laterally
+    for i, src_pos in enumerate(networkConfig["columnPositions"]):
+      suffixSrc = "_" + str(i)
+      for j, dest_pos in enumerate(networkConfig["columnPositions"]):
+        if i != j and numpy.linalg.norm(numpy.asarray(src_pos) -
+        numpy.asarray(dest_pos)) <= networkConfig["maxConnectionDistance"]:
```
scottpurdy: indent

htmresearch/frameworks/layers/l2_l4_network_creation.py
```diff
@@ -333,3 +359,74 @@ def createMultipleL4L2Columns(network, networkConfig):
...
+
+      layerConfig["numOtherCorticalColumns"] = numCorticalColumns - 1
+
+      suffix = "_" + str(i)
+      network = createL4L2Column(network, networkConfigCopy, suffix)
+
+    # Now connect the L2 columns laterally
+    for i, src_pos in enumerate(networkConfig["columnPositions"]):
+      suffixSrc = "_" + str(i)
+      for j, dest_pos in enumerate(networkConfig["columnPositions"]):
```
scottpurdy: Can we avoid iterating over all column positions and instead figure out the 2D square of potential columns that could be connected? As coded, this is `n^2` with n being number of columns where it could just be `n*c` with c being the `maxConnectionDistance` squared

MaxASchwarzer: This is hard to do given that we allow the user to specify arbitrary coordinates for each column, and given that it's only an O(n^2) look over a pretty small n I don't think it's worth a substantial refactor.  If we have enough columns to make this loop take a long time, our code will be slow no matter what.

tests/frameworks/layers/l2l4_network_creation_test.py
MaxASchwarzer: Done.  This was also an issue in some of the preexisting tests, and I've fixed it there as well.

PR #784, opened 2017-07-27
RES-502 Add online learning for column pooler

MaxASchwarzer:
This PR adds online learning capabilities to the column pooler, and changes our feedback experiments to use the apical dependent temporal memory. Online learning is disabled by default, for compatibility with our existing sensorimotor inference research.


htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -44,7 +44,8 @@ def __init__(self,
                apicalInputSize=0,
                cellsPerColumn=32,
                activationThreshold=13,
-               initialPermanence=0.21,
+               reducedBasalThreshold=10,
+               initialPermanence=0.61,
```
mrcslws: Let's use the same default `initialPermanence` as `ApicalTiebreakTemporalMemory` and `ApicalTMRegion`. The experiment code can override it to 0.61.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -115,10 +121,12 @@ def __init__(self,
     self.basalPredictedSegmentDecrement = basalPredictedSegmentDecrement
     self.apicalPredictedSegmentDecrement = apicalPredictedSegmentDecrement
     self.activationThreshold = activationThreshold
+    self.reducedBasalThreshold = 10
```
mrcslws: Use the provided threshold.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -174,34 +182,49 @@ def compute(self,
     Whether to grow / reinforce / punish synapses
     """
 
+    self.burstingColumns = []
```
mrcslws: This is the only mention of `self.burstingColumns`.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -174,34 +182,49 @@ def compute(self,
...
-      self.basalConnections.mapSegmentsToCells(activeBasalSegments),
-      self.apicalConnections.mapSegmentsToCells(activeApicalSegments))
+      matchingApicalSegments,
+      apicalPotentialOverlaps) = self._calculateSegmentActivity(
+        self.apicalConnections, apicalInput, self.connectedPermanence,
+        self.activationThreshold, self.minThreshold)
+
+    apicallySupportedCells = self.apicalConnections.mapSegmentsToCells(
+        activeApicalSegments)
+    if self.useFeedback:
```
mrcslws: I think it'd be reasonable to remove this `self.useFeedback` check and just use `_calculateBasalSegmentActivity`. It already checks whether `reducedBasalThreshold != activationThreshold`, so there's no real performance overhead.

So we can remove `self.useFeedback` and remove the `else` statement.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -174,34 +182,49 @@ def compute(self,
...
 
+      predictedCells = self.basalConnections.mapSegmentsToCells(activeBasalSegments)
+    self.predictedCells = predictedCells
     # Calculate active cells
     (correctPredictedCells,
      burstingColumns) = np2.setCompare(predictedCells, activeColumns,
                                        predictedCells / self.cellsPerColumn,
                                        rightMinusLeft=True)
+
+    self.inhibitionSignal = len(predictedCells) > self.activationThreshold
```
mrcslws: This is the only mention of `self.inhibitionSignal`.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -419,6 +442,57 @@ def _calculateSegmentActivity(connections, activeInput, connectedPermanence,
             matchingSegments,
             potentialOverlaps)
 
+  @staticmethod
+  def _calculateBasalSegmentActivity(connections, activeInput,
```
mrcslws: You could do all of this with a single `_calculateSegmentActivity` method. Just have a `reducedThresholdCells` parameter on the method. When we use it to calculate active apical segments, just pass in `()` for the `reducedThresholdCells`.

htmresearch/algorithms/apical_dependent_temporal_memory.py
mrcslws: Oops, I didn't realize this was actually a new feature. (We discussed offline -- maybe name it `disableApicalDependence`)

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
 
     @param  learn (bool)
             If True, we are learning a new object
-    """
 
+    @param predictedInput (sequence)
+           Sorted indices of active feedforward input bits
```
mrcslws: This parameter needs a good docstring. This made me think these were "predicted active cells", when really these cells aren't necessarily active (and this is vital to how the algorithm works.)

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+    if not learn:
+      self._computeInferenceMode(feedforwardInput, lateralInputs)
+
+    elif not self.onlineLearning:
       self._computeLearningMode(feedforwardInput, lateralInputs,
                                 feedforwardGrowthCandidates)
     else:
-      self._computeInferenceMode(feedforwardInput, lateralInputs)
+      if (predictedInput is not None and
+          len(predictedInput) > self.predictedInhibitionThreshold):
```
mrcslws: I'm surprised that this code isn't saying "Alright, there was a prediction, so I'm going to require all active cells in the pooling layer to have feedforward support if they want to stay active." Instead, if the input doesn't match the prediction, it's as if there's no feedforward input. So the inertia will keep many previously active cells active. Then we activate a few other random cells, then learn on that SDR?

For example, if the sequence ABCDE is known, and then you're seeing APQRS for the first time, upon "P" I'd expect this code to deactivate the "ABCDE" sequence SDR, then activate a random "APQRS" SDR. Instead, depending on the inertia and learningTolerance parameters, it will create a "APQRS" SDR that is similar to the "ABCDE" SDR?

(Aside: I don't know how a random "APQRS" SDR would ever learn "A". For object recognition, that's fine, because you can always revisit locations, but for sequence learning I don't know how you would ever prepend elements to a new sequence, other than by using a new form of learning that goes multiple timesteps backward.)

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+        # before actually attempting to learn.
+        self._computeInferenceMode(feedforwardInput, lateralInputs)
+        self._computeOnlineLearningMode(feedforwardInput, lateralInputs,
+                                        feedforwardGrowthCandidates)
+      else:
+        # If there isn't predicted input and we have a single SDR,
+        # we are extending that representation and should just learn.
+        self._computeOnlineLearningMode(feedforwardInput, lateralInputs,
+                                  feedforwardGrowthCandidates)
+        self._computeInferenceMode(feedforwardInput, lateralInputs)
```
mrcslws: Why do we infer here? `_computeOnlineLearningMode` should have already selected the active cells.

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+    @param  feedforwardGrowthCandidates (sequence or None)
+            Sorted indices of feedforward input bits that the active cells may
+            grow new synapses to.  This is assumed to be the predicted active
+            cells of the input layer.
+    """
+    prevActiveCells = self.activeCells
+    cellsToLearn = set(self.activeCells)
+
+    # If there are not enough previously active cells, select random subset of
+    # cells to learn on.  If there were only some active cells, we assume that
```
mrcslws: This comment makes it seem like you're selecting a random subset of the previously active cells.

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+    # we are learning an object related to what is represented by those cells;
+    # if there were none, we are learning a new object.
+    # This case is the only way different object representations are created.
+    if len(cellsToLearn) < self.sdrSize:
+      if len(cellsToLearn) < self.sdrSize*(1 - self.learningTolerance):
```
mrcslws: I think this should be all-or-none. Either use the current set of active cells, or clear them and activate a random SDR. We shouldn't ever take the current active cells and add a few random cells. That creates a totally different type of SDR where the bits have semantic meaning. That type of SDR doesn't have the union property. (Also, it seems wrong to have the learning code activate *random* cells immediately after the inference code just chose not to activate some cells that had lateral / feedforward support).

Put differently: the learning code should either reset the active cells, or it should leave them alone. Nothing in-between.

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+        cellsToLearn = set()
+      numToAdd = self.sdrSize - len(cellsToLearn)
+      newCells = _sampleRange(self._random,
+                              0, self.numberOfCells(),
+                              step=1, k=numToAdd)
+      cellsToLearn |= set(newCells)
```
mrcslws: It would be best to do this all with numpy arrays. Sets are slower because every element of the set can be of any type, while every element of a numpy array is of a single type, so the underlying C code can work very quickly.

~~~python
cellsToLearn = np.union1d(cellsToLearn, newCells)
~~~

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+
+      cellsToLearn = numpy.asarray(list(cellsToLearn))
+      cellsToLearn.sort()
+      self.activeCells = numpy.union1d(self.activeCells, cellsToLearn)
```
mrcslws: It's weird that this line of code happens again further down. I think you can arrange this so that this union and sort only happen once.

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+    if len(feedforwardInput) > 0:
+      self._learn(self.proximalPermanences, self._random,
+                  cellsToLearn, feedforwardInput,
+                  feedforwardGrowthCandidates, self.sampleSizeProximal,
+                  self.initialProximalPermanence, self.synPermProximalInc,
+                  self.synPermProximalDec, self.connectedPermanenceProximal)
+
+      # External distal learning
+      # We should do this no matter what, since other columns might still be
+      # learning useful things even if we're not.
```
mrcslws: Is this comment saying that we should do this even when there's no feedforward input? (That's not what the code is doing.)

htmresearch/algorithms/column_pooler.py
```diff
@@ -174,16 +192,127 @@ def compute(self, feedforwardInput=(), lateralInputs=(),
...
+      cellsToLearn |= set(newCells)
+
+      cellsToLearn = numpy.asarray(list(cellsToLearn))
+      cellsToLearn.sort()
+      self.activeCells = numpy.union1d(self.activeCells, cellsToLearn)
+      self.activeCells.sort()
+
+      # Form initial internal distal links for the new SDR.  This is crucial in
+      # online learning.  Without it, the SDR may fail to persist.
+      self._learn(self.internalDistalPermanences, self._random,
```
mrcslws: It's weird that we have this piece of learning that happens only once ever for this particular SDR. All these other horizontal connections learn continuously, but the ones internal to the layer learn once. We essentially aren't modeling permanences for these connections -- we're assuming that the initial permanence is above the connected permanence and then never updating it.

Maybe instead, when you're in the "else" statement where you're extending the sequence, you should allow previously active cells to stay active even if they don't have lateral support. Then I think you can learn these synapses alongside all the other lateral connections. It will work even with a low "initialDistalPermanence".

htmresearch/algorithms/column_pooler.py
MaxASchwarzer: That's correct, it could never learn "A", unless you were to replay the sequence with the "APQRS" SDR still active.  Fundamentally, I think this is an intractable problem without either backtracking or a supervisory signal, and it's outside of the scope of this to have either.

htmresearch/algorithms/column_pooler.py
MaxASchwarzer: This "learning tolerance" is necessary because of the possibility of errors, basically.  It's not guaranteed that all cells will become active when they should, and it is quite possible for an extra cell or two to become active.  If we insisted on having _exactly_ the right number of cells active, we wipe the SDR far more often than is desirable, and end up relearning things that should have been grouped together.  This tends to lead to results like this, where the redundant representations harm performance noticeably:
prediction performance with swapped stimuli no shared range

MaxASchwarzer: This is a good point.  The second if-statement should be removed.

MaxASchwarzer:
@mrcslws I've now addressed everything you raised, largely by cleaning up the code.  In general, I'd like to note that the column pooler online learning, as implemented here, is actually designed to create symbolic SDRs rather than semantic ones.  The code has been simplified greatly to make that more apparent.  As it happens, I am running experiments focused on creating semantic SDRs using the union pooler, but that's in a different branch.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -386,9 +408,16 @@ def _calculateLearning(self,
 
   @staticmethod
   def _calculateSegmentActivity(connections, activeInput, connectedPermanence,
-                                activationThreshold, minThreshold):
+                                activationThreshold, minThreshold,
+                                reducedBasalThreshold,
+                                reducedBasalThresholdCells = ()):
```
mrcslws: These could just be `reducedThreshold` and `reducedThresholdCells`. This method should never mention "apical" or "basal" -- it can be used for both. If someone wanted, they could use this method as-is to reduce the apical threshold for a set of cells.

htmresearch/algorithms/apical_dependent_temporal_memory.py
```diff
@@ -386,9 +408,16 @@ def _calculateLearning(self,
...
     """
-    Calculate the active and matching segments for this timestep.
+    Calculate the active and matching basal segments for this timestep.
+
+    The difference with _calculateApicalSegmentActivity is that cells
+    with active apical segments (collected in reducedBasalThresholdCells) have
+    a lower activation threshold for their basal segments (set by
+    reducedBasalThreshold parameter).
```
mrcslws: You could just change this docstring back to "Calculate the active and matching segments for this timestep."

htmresearch/algorithms/column_pooler.py
```diff
@@ -207,43 +264,53 @@ def _computeLearningMode(self, feedforwardInput, lateralInputs,
...
-      if len(prevActiveCells) > 0:
-        self._learn(self.internalDistalPermanences, self._random,
-                    self.activeCells, prevActiveCells, prevActiveCells,
-                    self.sampleSizeDistal, self.initialDistalPermanence,
-                    self.synPermDistalInc, self.synPermDistalDec,
-                    self.connectedPermanenceDistal)
-
       # External distal learning
+      # We should do this no matter what, since other columns might still be
+      # learning useful things even if we're not.
```
MaxASchwarzer: Sorry, this has gotten bounced back and forth several times.  I corrected the previous version to learn constantly, but then it had to be changed back when I merged the two learning methods, and I forgot to update the comment.

htmresearch/algorithms/column_pooler.py
```diff
@@ -207,43 +264,53 @@ def _computeLearningMode(self, feedforwardInput, lateralInputs,
...
       for i, lateralInput in enumerate(lateralInputs):
         self._learn(self.distalPermanences[i], self._random,
                     self.activeCells, lateralInput, lateralInput,
                     self.sampleSizeDistal, self.initialDistalPermanence,
                     self.synPermDistalInc, self.synPermDistalDec,
                     self.connectedPermanenceDistal)
 
+      # Internal distal learning
+      self._learn(self.internalDistalPermanences, self._random,
+                  self.activeCells, prevActiveCells, prevActiveCells,
```
MaxASchwarzer: Yeah, this bothered me a lot, but our current tests are dependent on this implementation, so I think it's best to leave it as it is for now.

htmresearch/algorithms/column_pooler.py
```diff
@@ -69,6 +72,22 @@ def __init__(self,
     @param  sdrSize (int)
             The number of active cells in an object SDR
 
+    @param  onlineLearning (Bool)
+            Whether or not the column pooler should learn in online mode.
+
+    @param  learningTolerance (float)
```
MaxASchwarzer: I like this from a simplicity standpoint, but it will be important to make sure that maxSdrSize and minSdrSize default to sdrSize, no matter what value for that parameter was passed in.  Otherwise, a user could easily break everything by passing in sdrSize = 20 and not passing a minSdrSize, which would default to 40.  I think it's easy to deal with this by making their defaults None, and just setting them to sdrSize if None during initialization.

MaxASchwarzer:
Tests all pass, and the old graphs look good, so I've merged this.  It's a good change, and it fixes a lot of the core issues we've had, but we're not all the way to true continuous learning yet.

PR #791, opened 2017-08-29
Set random seed in simple object machine

ywcui1990:
@subutai @lscheinkman Please review

I found that the location pool and the feature pool are different every time, even though the (feature, location) pair indices are the same.


lscheinkman:
👍 
@ywcui1990 This PR fixed the random issue i was having today. Thanks

PR #797, opened 2017-09-06
Update experiment descriptions for TM

ywcui1990:
@lscheinkman The experiment results for the Neural Computation paper can be reproduced following the instruction in the README.md. Please review

PR #800, opened 2017-09-11
Replaces #797

lscheinkman:
This PR replaces PR #797 removing requirements from global requirements.txt (b5a9c71)
@subutai Please review


projects/sequence_prediction/continuous_sequence/run_arima.R
```diff
@@ -4,11 +4,11 @@ library(forecast)
 library(TSPred)
 
 # set working directory here
-setwd('/Users/ycui/nta/nupic.research/projects/sequence_prediction/continuous_sequence')
+setwd('/Users/ycui/nta/htmresearch/projects/sequence_prediction/continuous_sequence')
```
subutai: This might need to be changed?

lscheinkman: I did not validate experiments written in  R or matlab with this PR, only experiments written in python. I am not sure if those experiments still work or the software requirements. I will probably validate in a future PR.

subutai: Was just thinking that `setwd('/Users/ycui/nta/...` will work on very few laptops 😄

subutai: But fine to merge...

lscheinkman: How about this? 00a9eb1

It uses the current script location instead of hardcoded location

PR #811, opened 2017-09-25
Branch res 617

vincentliuk:
remove all the places with parameter distalSegmentInhibitionFactor


htmresearch/algorithms/column_pooler.py
```diff
@@ -26,7 +26,7 @@
 
 
 class ColumnPooler(object):
-  """
+  """,
```
subutai: Extra comma?

vincentliuk: comma deleted, and committed again

PR #833, opened 2017-12-22
New Lateral Pooler Implementation

mirkoklukas:
- Rewrote the Lateral Pooler class: It does now inherit from the SP, and extends it. 
- Changed the other files accordingly


htmresearch/support/lateral_pooler/datasets.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+import numpy as np
+from scalar_encoder import ScalarEncoder 
+from scipy.io import loadmat
+from sklearn.datasets import fetch_mldata
```
scottpurdy: This code won't run by default if the libraries aren't in the projects requirements.txt file. Fine to leave them out if the lateral pooler is still experimental but you might want to indicate that in the module doc string in that file.

PR #834, opened 2018-01-18
patternSimilarityDetection

vincentliuk:
population of neurons' firing pattern similarity detection with similar movie stimulus.


vincentliuk:
@subutai here are the new codes I have been working on to extract the firing pattern of population of neurons corresponding to the similar movie frame stimulus.

vincentliuk:
merged done

PR #836, opened 2018-01-22
sparsity+src

vincentliuk:
sparsity code with several src codes


vincentliuk:
@subutai  Here are the python codes for sparsity and several other source codes. I have tested it on jupyter notebooks. However, there might be some data format difference which could be solved manually.

subutai:
@lscheinkman Can you take a look? Thanks.

projects/high_order_cell_assembly/get_response_mat.py
```diff
@@ -0,0 +1,42 @@
+def get_response_mat(spiketrain, imgPara, stimType, goodCells, plotRaster):
```
lscheinkman: Need to import  `numpy`

projects/high_order_cell_assembly/sparsity_over_trial.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
+import scipy.io as sio 
+import matplotlib.pyplot as plt
+import get_response_mat
+import calculate_sparsity_over_trials
```
lscheinkman: Replace
```
import get_response_mat
import calculate_sparsity_over_trials
```
With
``` 
from get_response_mat import get_response_mat
from calculate_sparsity_over_trials import calculate_sparsity_over_trials
```

lscheinkman:
Calling `python sparsity_over_trial.py` from command line does not work. Here is the error:
```
Traceback (most recent call last):
  File "sparsity_over_trial.py", line 24, in <module>
    stim_type = imgPara['stim_type']
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

Can you add the notebook you used to test?

lscheinkman:
Replaced by #838

PR #859, opened 2018-06-01
Add code for location layer noise experiments

MaxASchwarzer:
This PR adds code for the noise experiments I have been conducting over the past two weeks.  It also implements provisions for other types of location layer functionality, including the four-corners method and a purely discrete version of our location layer.


htmresearch/algorithms/location_modules.py
```diff
@@ -184,6 +200,10 @@ def activateRandomLocation(self):
     Set the location to a random point.
     """
     self.activePhases = np.array([np.random.random(2)])
+    if self.anchoringMethod == "discrete":
+      # Need to place the phase in the middle of a cell
+      self.active_phases = np.floor(
+      self.activePhases * self.cellDimensions)/self.cellDimensions
```
scottpurdy: indentation is off here

also you have `self.active_phases` being set using `self.activePhases`. Is that a typo? Regardless, use the camelcase version and change the name of the other to clearly distinguish it if they aren't supposed to be the same

htmresearch/frameworks/location/path_integration_union_narrowing.py
```diff
@@ -262,24 +264,36 @@ def learnObject(self, objectDescription):
     self.reset()
     self.column.activateRandomLocation()
 
-    for iFeature, feature in enumerate(objectDescription["features"]):
-      self._move(feature)
+    if randomLocation or useNoise:
+      numIters = 5
```
scottpurdy: This would be cleaner as a passed in argument that defaults to 1 but no need to change for this PR

projects/union_path_integration/convergence_simulation.py
```diff
@@ -117,6 +116,14 @@ def doExperiment(cellDimensions,
 
   thresholds = numModules
   perModRange = float(90.0 / float(numModules))
+
+  # No need to keep track of extra points for corners and discrete methods.
+  if anchoringMethod == "corners":
+    cellCoordinateOffsets = (.0001, .5, .9999)
```
scottpurdy: Don't need `.5` do you?

scottpurdy: Also might want to add a note somewhere to indicate that the cellCoordinateOffsets don't get used if using corners or discrete.

PR #870, opened 2018-08-10
Add model that more closely mimics rat entorhinal modules

mrcslws:
Here's what distinguishes this model from the other:
- It separately models the "bump size" and the "readout resolution" rather than using the 1/cellsPerModuleAxis for both.
- It models how a union of bumps is harder to readout than a single bump. (In this approach, more cells become "active" than just the union of the active cells)
- Hexagonal grids. The firing fields of a cell are arranged in a hexagon rather than in a square.
- Each firing field is round instead of square.

Other changes:

- Fix a bug: PIUNExperiment.representationSet wasn't being properly filled, so we weren't correctly detecting temporary convergences on a wrong object.
- Stop seeding the object generator with the number of objects. It should generate different objects on different runs.
- Add couple other capabilities to convergence simulation.
- Tweak the Superficial2DLocationModule so that it shares a more common interface with the new ThresholdedGaussian2DLocationModule.

### DIFF
--- a/htmresearch/algorithms/location_modules.py
+++ b/htmresearch/algorithms/location_modules.py
@@ -32,6 +32,448 @@ from htmresearch.algorithms.multiconnections import Multiconnections
 from nupic.bindings.math import SparseMatrixConnections, Random
 
 
+class ThresholdedGaussian2DLocationModule(object):
+  """
+  A model of a grid cell module. The module has one or more Gaussian activity
+  bumps that move as the population receives motor input. When two bumps are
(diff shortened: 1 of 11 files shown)


scottpurdy:
What do you think about creating a copy of `convergence_simulation.py` specific to the new method? I think it will be hard to keep track of all the command line options if we make convergence_simulation handle all the different location module types / options.

mrcslws:
@scottpurdy good point. I made the change. And then I squashed the commits to avoid unnecessary changes in the git-blame.

PR #881, opened 2018-09-17
PUB-867: Location network framework

lscheinkman:
@subutai Please review.
There are 2 tests for now, one for the region itself (`tests/regions/location_region_test.py`) and one for the `createL4L6aLocationColumn` network, which mimics the `projects/union_path_integration/infer_hand_crafted_objects.py` experiment using the network API.

Currently there is no tests for `createL246aLocationColumn`.

### DIFF
--- a/htmresearch/regions/Guassian2DLocationRegion.py
+++ b/htmresearch/regions/Guassian2DLocationRegion.py
@@ -257,7 +257,7 @@ class Guassian2DLocationRegion(PyRegion):
           defaultValue="probabilistic",
           count=0,
         ),
-        learn=dict(
+        learningMode=dict(
           description="A boolean flag that indicates whether or not we should "
                       "learn by associating the location with the sensory "
                       "input",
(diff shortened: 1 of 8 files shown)


htmresearch/regions/Guassian2DLocationRegion.py
```diff
@@ -0,0 +1,445 @@
...
+    PyRegion.__init__(self, **kwargs)
+
+  def initialize(self):
+    """ Initialize grid cell modules """
+
+    if self._modules is None:
+      self._modules = []
+      for i in xrange(self.moduleCount):
+        self._modules.append(ThresholdedGaussian2DLocationModule(
+          self.cellsPerAxis,
```
subutai: Should we make this call use keyword parameters in case the order changes later? I.e.cellsPerAxis = self.cellsPerAxis, etc.?

htmresearch/regions/Guassian2DLocationRegion.py
```diff
@@ -0,0 +1,445 @@
...
+        bumpOverlapMethod=dict(
+          description="Specifies the firing rate of a cell when it's part of "
+                      "two bumps. ('probabilistic' or 'sum')",
+          dataType="Byte",
+          accessMode="Read",
+          constraints=("enum: probabilistic, sum"),
+          defaultValue="probabilistic",
+          count=0,
+        ),
+        learn=dict(
```
subutai: `learningMode`?

PR #886, opened 2018-09-26
Don't use difference between phases as a distance

mrcslws:
This changes the experiment to calculate distances on a rhombus, not on a square. It also changes the experiment to use a basis that always has a hexagonal grid on some plane in k-dimensional space.


mrcslws:
Dependent on https://github.com/numenta/htmresearch-core/pull/61

PR #902, opened 2018-11-14
RES-789: Compute L2-L4-L6 accuracy using YCB/Thing data

lscheinkman:
@subutai Please review. Specially the 3D projection commit 7447c59


htmresearch/regions/GridCellLocationRegion.py
```diff
@@ -315,11 +322,14 @@ def __init__(self,
                learningMode=False,
                seed=42,
                dualPhase=True,
+               dimensions=2,
```
subutai: Why is dimensions=2 for the seed?

PR #912, opened 2019-01-04
Initial pytorch experiments

subutai:


subutai:
@lscheinkman - this version has a working CNN. Still finishing up tests though, so might find more bugs.

PR #914, opened 2019-01-08
Improved CNN code

subutai:


htmresearch/frameworks/pytorch/sparse_mnist_cnn.py
```diff
@@ -71,7 +71,8 @@ def __init__(self,
                c1OutChannels=20,
                c1k=20,
                n=50,
-               useDropout=True,
+               k=50,
```
lscheinkman: Missing doc for `k`. I am assuming it is the same as `l2k`

subutai: Thanks - fixed the docstring.

PR #927, opened 2019-01-17
RES-817: weight trimming

lscheinkman:
Use `min_weight` configuration parameter to trim the weights whose absolute magnitude is <= `min_weight`

### DIFF
--- a/htmresearch/frameworks/pytorch/mnist_sparse_experiment.py
+++ b/htmresearch/frameworks/pytorch/mnist_sparse_experiment.py
@@ -396,17 +396,9 @@ class MNISTSparseExperiment(PyExperimentSuite):
 
   def pruneWeights(self, params):
     """
-    Prune the weights whose absolute magnitude is <= params["min_weight"]
+    Prune the weights whose absolute magnitude is < params["min_weight"]
     """
-    minWeight = params.get("min_weight", 0.0)
-
-    # Collect all weights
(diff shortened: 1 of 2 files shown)


htmresearch/frameworks/pytorch/mnist_sparse_experiment.py
```diff
@@ -393,10 +396,17 @@ def runNoiseTests(self, params):
 
   def pruneWeights(self, params):
     """
-    TODO: Prune the weights whose absolute magnitude is <= params["minWeight"]
+    Prune the weights whose absolute magnitude is <= params["min_weight"]
```
subutai: I think this should say "< params["min_weight"]" 

In other words, it prunes stuff strictly less than min_weight, which is actually better. min_weight of 0 will have no effect.

htmresearch/frameworks/pytorch/mnist_sparse_experiment.py
```diff
@@ -221,8 +221,11 @@ def iterate(self, params, repetition, iteration):
 
   def finalize(self, params, rep):
     """
-    Save the full model once we are done.
+    Called once we are done.
     """
+    self.pruneWeights(params)
```
subutai: Could you make pruneWeights a method of the model itself? I might want to save the unpruned model, and then separately prune it for analysis.

PR #929, opened 2019-01-17
Batchnorm and switch to sparse_net

subutai:


subutai:
@lscheinkman 
Implemented batchNorm - it seems to help everywhere, so made it default to True. 
Switched to sparse_net
Also changed a few small things.

htmresearch/frameworks/pytorch/sparse_speech_experiment.py
```diff
@@ -177,9 +217,13 @@ def createLearningRateScheduler(self, params, optimizer):
     if lr_scheduler is None:
       return None
 
-    lr_scheduler_params = params.get("lr_scheduler_params", None)
```
lscheinkman: I guess i need to rethink the way we pass parameters to the `lr_scheduler`. This logic should be handled by the `expsuite`. I will take a look and see if i can fix it there

htmresearch/frameworks/pytorch/k_winners.py
```diff
@@ -154,8 +154,10 @@ def updateDutyCycleCNN(x, dutyCycle, dutyCyclePeriod, learningIterations):
   dutyCycle.reshape(-1).add_(s)
   dutyCycle.div_(period)
 
+  # This seems to happen, but very rarely. Keep an eye on it.
```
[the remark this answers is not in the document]
subutai: I don't think I saw it before. It happened after, during one minibatch.

htmresearch/frameworks/pytorch/sparse_speech_experiment.py
subutai: Thanks.

PR #936, opened 2019-01-22
Added speech noise experiments plus other tweaks

subutai:


projects/speech_commands/run_experiment.py
```diff
@@ -40,9 +43,24 @@
   projectDir = path.dirname(suite.options.config)
   dataDir = path.join(path.abspath(projectDir), "data")
 
-  if not path.isdir(path.join(dataDir, "speech_commands")):
-    tar = tarfile.open(path.join(dataDir, "speech_commands.tar.gz"))
-    tar.extractall(path=dataDir)
-    tar.close()
+  if "DOMINO_WORKING_DIR" in os.environ:
```
lscheinkman: Not for tomorrow, but should we use the 'domino' dataset instead of this code?

projects/speech_commands/experiments.cfg
```diff
@@ -77,23 +77,51 @@ model_type = "linear"
...
+learning_rate = 0.04
+momentum = 0.9
+weight_sparsity = 1.0
+dropout = 0
+log_interval = 40
+test_noise_every_epoch = False
+batches_in_epoch = 214
+batch_size = 96
+model_type = "resnet9"
+seed = [51, 52]
```
lscheinkman: This is not necessary. The seed is updated on the reset based on the repetition.
https://github.com/numenta/htmresearch/blob/master/htmresearch/frameworks/pytorch/mnist_sparse_experiment.py#L65

subutai: Good point. I'll fix these after Wed 😄

projects/speech_commands/run_experiment.py
subutai: Yes.

PR #938, opened 2019-01-23
Fallback to matplotlib when plotly is not available

lscheinkman:
@subutai Please review. See attached plot
[effect_of_n.pdf](https://github.com/numenta/htmresearch/files/2785165/effect_of_n.pdf)


subutai:
Looks great. Can you just get rid of the plotly code?

PR #944, opened 2019-01-24
Fix gpu and default params

lscheinkman:
Added missing default values for the `experiments_paper.cfg` and fixed `daemonic processes are not allowed to have children` error when using GPU.

@subutai I was able to run all the 8 experiments in experiments_paper.cfg in 20 minutes after this fix using domino's 
`GPU g3.4xlarge (CUDA 9)` machine combined with `htmresearch dev (GPU Tools CUDA9)`

### DIFF
--- a/projects/sdr_paper/pytorch_experiments/experiments_paper.cfg
+++ b/projects/sdr_paper/pytorch_experiments/experiments_paper.cfg
@@ -21,6 +21,9 @@ create_plots = False
 count_nonzeros = True
 test_noise_every_epoch = True
 
+use_batch_norm = True
+boost_strength_factor = 1.0
+optimizer = SGD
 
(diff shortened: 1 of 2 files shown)


subutai:
Excellent!

projects/sdr_paper/pytorch_experiments/experiments_paper.cfg
```diff
@@ -21,6 +21,9 @@ create_plots = False
 count_nonzeros = True
 test_noise_every_epoch = True
 
+use_batch_norm = True
```
subutai: should we default to False for MNIST since batch norm didn't help?

lscheinkman: yes

PR #957, opened 2019-02-06
Trained RNN localization model.

maxwellpollack:


subutai:
We don't normally checkin model files, like trained_model.pt - it's fine to have them locally but we don't want them in our repository. Could you take it out of the commit list?

maxwellpollack:
Donezo!

PR #958, opened 2019-02-07
Create a new pre-process speech commands dataset with all transformatons applied

lscheinkman:


projects/speech_commands/run_experiment.py
```diff
@@ -38,29 +38,29 @@
...
-      tar.close()
-    else:
-      print("Apparently this still exists:", path.join(dataDir, "speech_commands"))
-
-    print("listing files:")
-    files = os.listdir(path.join(dataDir, "speech_commands"))
-    print(files)
-    files = os.listdir(path.join(dataDir, "speech_commands", "train", "one"))
-    print("train/one", len(files), files[0:10])
+  #
```
subutai: Should we get rid of these commented lines?

PR #959, opened 2019-02-08
 RES-809: Compute weight pruning on the speech commands experiments

lscheinkman:


projects/speech_commands/analyze_nonzero.py
```diff
@@ -0,0 +1,256 @@
+# ----------------------------------------------------------------------
+# Numenta Platform for Intelligent Computing (NuPIC)
+# Copyright (C) 2018, Numenta, Inc.  Unless you have an agreement
```
subutai: 2019 😄

PR #965, opened 2019-02-15
RES-855: Add option to normalize weights intialization to the number of non-zeros

lscheinkman:
@subutai Added the optional parameter `normalize_weights` to the `experiments.cfg' controlling this functionality. Default to `False`

### DIFF
--- a/htmresearch/frameworks/pytorch/linear_sdr.py
+++ b/htmresearch/frameworks/pytorch/linear_sdr.py
@@ -20,6 +20,7 @@
 # ----------------------------------------------------------------------
 
 from __future__ import print_function
+import math
 import numpy as np
 
 import torch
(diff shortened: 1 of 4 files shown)


htmresearch/frameworks/pytorch/linear_sdr.py
```diff
@@ -108,6 +114,24 @@ def __init__(self,
...
+  def normalizeWeights(self):
+    """
+    Initialize the weights using kaiming_uniform initialization normalized to
+    the number of non-zeros in the layer instead of the whole input size.
+    Similar to torch.nn.Linear.reset_parameters() but applying weight sparsity
+    to the input size
+    """
+    _, inputSize = self.l1.weight.shape
+    fan = int(inputSize * self.weightSparsity)
+    gain = nn.init.calculate_gain('leaky_relu', math.sqrt(5))
```
subutai: Do you know what this gain ends up being usually?

lscheinkman: Here, using 'leaky_relu'  gain is always `math.sqrt(2/6)`
See https://pytorch.org/docs/stable/_modules/torch/nn/init.html

PR #967, opened 2019-03-08
RES-863:  Use 'Layer Norm' instead of 'Batch Norm' on linear networks

lscheinkman:
@subutai Please review.
The main change in this PR is the switch from `BatchNorm` to `LayerNorm` on `LinearSDR` module.
See https://arxiv.org/abs/1607.06450


subutai:
This doesn’t affect any of our existing sparse networks right?

lscheinkman:
If affects `LinearSDR`. We can now use batch_size=1

lscheinkman:
However we cannot user `LayerNorm` on `CNNSDR2d`. It is only good for Linear and RNN networks

htmresearch/frameworks/pytorch/cnn_sdr.py
```diff
@@ -124,6 +129,8 @@ def __init__(self,
     shape = self.outputSize()
```
subutai: Don't you have to change outputSize() too?

htmresearch/frameworks/pytorch/linear_sdr.py
```diff
@@ -81,10 +81,12 @@ def __init__(self,
     self.l1 = nn.Linear(inputFeatures, self.n)
     self.weightSparsity = weightSparsity
     self.learningIterations = 0
+    if self.k <= 0:
+      self.k = self.n
 
     self.bn = None
     if useBatchNorm:
-      self.bn = nn.BatchNorm1d(self.n, affine=False)
+      self.bn = nn.LayerNorm(self.n)
```
subutai: How would this affect GSC results?

htmresearch/frameworks/pytorch/linear_sdr.py
```diff
@@ -81,10 +81,12 @@ def __init__(self,
     self.l1 = nn.Linear(inputFeatures, self.n)
     self.weightSparsity = weightSparsity
     self.learningIterations = 0
+    if self.k <= 0:
+      self.k = self.n
 
     self.bn = None
     if useBatchNorm:
```
subutai: Should probably change the name of the parameter too.

htmresearch/frameworks/pytorch/cnn_sdr.py
lscheinkman: `outputSize` was already using stride parameter

PR #976, opened 2019-03-26
RES-894: Baseline continual learning experiment

lscheinkman:
@subutai  Please review
This is just the baseline for continual learning experiment. You can see the results of the sparseCNN2 vs denseCNN2 in the [projects/continual_learning/mnist/plots.ipynb](https://github.com/numenta/htmresearch/blob/5361e4be28d44050b9ba67aa386419673ef0e645/projects/continual_learning/mnist/plots.ipynb)
Here is the main commit: d849d30


htmresearch/frameworks/pytorch/modules/flatten.py
```diff
@@ -0,0 +1,30 @@
...
+# You should have received a copy of the GNU Affero Public License
+# along with this program.  If not, see http://www.gnu.org/licenses.
+#
+# http://numenta.org/licenses/
+# ----------------------------------------------------------------------
+import torch.nn as nn
+
+
+
+class Flatten(nn.Module):
```
subutai: I'm surprised this isn't already in pytorch!

lscheinkman: it is in tensorflow

PR #977, opened 2019-03-26
Updates for paper

subutai:


subutai:
@lscheinkman This was not aggregating scores correctly

PR #987, opened 2019-05-05
Minor rewording of README

TOertel:


subutai:
Haha good catch @TOertel ! 

@rhyolight Can you verify the CLA has been signed before merging? Unfortunately we need it even for these small changes.

